PR #8, opened 2016-12-17
Improve docker

linuxbandit:
Docker provisioning happens "the right way"
everything works from beginning to end


linuxbandit:
wrong branch

PR #11, opened 2017-01-24
Forward MongoDB port to the host machine

serge1peshcoff:
To have the ability to use 'localhost:27017' instead of '172.18.0.2:27017' (or something like that) to connect to MongoDB.


linuxbandit:
good one, I am actually surprised the port is not already forwarded but not even exposed

PR #14, opened 2017-03-28
Traefik

blacksph3re:
Added traefik as a proxy so we no longer have to expose several ports


linuxbandit:
So as I said on slack, looks good - however now I have returned so I can actually try it out

Only question I have is, why did the PR come from the branch AEGEE:oms-docker/traefik instead of your own repo? Why on master and not on dev?

i3anaan:
I agree with Fabrizio, this should be merged to dev. Closing because of this.

PR #16, opened 2017-04-18
Improve "Add Traefik as API gateway/reverse proxy"

linuxbandit:
Add repos for submodules that were forgotten :D
Make it work for Vagrant as well (README)
Add utility makefile to not type always same things

@i3anaan you are assigned because it is better to test also with Vagrant env (since nico and sergey still use pure docker)


linuxbandit:
Once it is ok, then @i3anaan can PR it into dev

docker/.editorconfig
```diff
@@ -3,10 +3,15 @@ root = true
 [*]
 end_of_line = lf
 insert_final_newline = true
-
-[*]
+indent_style = space
```
i3anaan, line 6: Should this file (docker/.editorconfig) be on git? Looks like some local config of the editor you are using :)

docker/Makefile
```diff
@@ -0,0 +1,30 @@
+  
+default:  
+	echo 'Options are bootstrap, start, monitor, stop, restart, nuke, clean (cleans untagged/unnamed images)'
+
+start:  
+	export COMPOSE_PROJECT_NAME=myaegee && sudo docker-compose up -d
+  
+bootstrap: start monitor
```
i3anaan, line 8: Would be nice to have consistent formatting here, I suggest putting the command on the next line (as done with the longer commands)

vagrant-post-script/orchestrate_docker.sh
```diff
@@ -1,8 +1,4 @@
 #!/bin/bash
 
 cd oms-docker/docker
-docker-compose up -d
-echo "############# ok until here"
-docker-compose logs -f omscore-bootstrap
-docker-compose logs -f omsevents-bootstrap
-echo "############# done"
+make bootstrap
```
i3anaan, line 4: When all vagrant does is call 'make bootstrap' is it really still useful to have Vagrant?

i3anaan:
Got this during Vagrant installation:

The process "php artisan optimize" exceeded the timeout of 300 seconds.  
Not sure how severe that is.

i3anaan:
There are actually multiple errors occurring during installation. Could be looked into, not sure if it should prevent merging.

linuxbandit:
I think the first one chained the others, but sure enough the PR won't be merged until they work properly. I just need to add a timeout

On a related note, of course you could try "docker-only"

i3anaan:
Using vagrant the installation did not succeed. It exited with the following line:
==> appserver: [36momsevents-bootstrap_1       |[0m {"code":"InternalError","message":"Could not register microservice, core communication failed"}Bootstrap finished

Full log can be found [here](https://gist.github.com/i3anaan/b566d7aea93c20f984b4c6ec8c7842cd)

Have fun solving! :D

i3anaan (changes requested):

docker/Makefile
linuxbandit: never mess with Makefile! works only that way

vagrant-post-script/orchestrate_docker.sh
linuxbandit: The Vagrant machine is just a standard dev env for everyone - just an ubuntu. Mainly for windows and mac people, but linux people should also use it only to avoid situations of "works on mine only" because that particular version of that particular library of that moment in time when it was installed was compiled in a certain way.. that's the rationale. 

Also soon the Vagrantfile will contain 2 VMs to swarm it

docker/.editorconfig
linuxbandit: it is to standardise code style across all developers - e.g. Makefile only accept tabs as indentation, although I personally like spaces. With Atom's editorconfig plugin, this file is read automatically and the setting is applied (otherwise Atom would convert it to spaces)

linuxbandit:
Thanks, it seems something with file permissions, very easy to solve. Don't try docker only as it should have same results

vagrant-post-script/orchestrate_docker.sh
blacksph3re: but docker already is an encapsulation to avoid "works on mine only"? The only thing that is outside of it is your makefile, and if that is broken on some computers you can still type those few commands by hand?

docker/Makefile
```diff
@@ -0,0 +1,30 @@
...
+stop:  
+	sudo docker-compose down
+  
+restart: stop bootstrap
+
+nuke:  
+	sudo docker-compose down -v
+  
+clean:
+	sudo docker rmi $(docker images -qf "dangling=true")
```
blacksph3re, line 22: That will remove all dangling images, also from completely unrelated docker projects. I don't think this should be in this makefile

docker/.editorconfig
i3anaan: While I am all up for standardizing, I am not sure if I like these kinds of files being in the repo. Curious to hear the opinion of others on this matter.

docker/Makefile
linuxbandit: (to give detail: on the same line there are "targets", on the different lines there are custom scripts that a target has to execute)

vagrant-post-script/orchestrate_docker.sh
linuxbandit: True, but again it standardises for mac and windows - also it could model the future infrastructure (currently it is an ubuntu box which is same version that is going to be on titan), and the main difference between hosts is the docker daemon which does influence a bit! For the company I work for I created the same setup (pick between vagrant+ or just docker) and pure docker didn't run smoothly

And of course the possibility to create a cluster on some machines with minimal setup

(Finally true about Make, but it is present by default on most distros, even arch :) )

docker/Makefile
linuxbandit: Thanks, I only thought of vagrant - good call

docker/Makefile
i3anaan: Okay, I trust you then :)

vagrant-post-script/orchestrate_docker.sh
i3anaan: Id say, since we already pretty much have it and it is not a lot of effort we can for sure keep it. If it proves troublesome to maintain in the future we can reconsider if we need to offer vagrant.

linuxbandit: I don't think it is hard to maintain and it has anyway a broader scope: you can tell any illiterate to install virtualbox and vagrant and to download the zip, and they can check the intranet on their laptop -> useful for demos

But that's beyond the scope of this comment, to get back in topic I could say: look [here](https://github.com/linuxbandit/oms-docker/blob/feat-swarm_ready/docker/docker-compose.yml#L311) and notice that a full path is needed to mount the code (~ doesn't work). Clearly vagrant standardises it. It is just faster to get one started without hassle, because I agree with nico that images are immutable anyway and give us enough stability (except the docker daemon being critical)

linuxbandit: Actually another thing is the filesystem drivers: does not depend on the docker daemon but on the OS.. sometimes they can bitch (I think that's the issue that my company has gotten into)

i3anaan: Okay, I think you made your point and I understand the added value of vagrant :)

docker/omsapplications-frontend/Dockerfile
```diff
@@ -1,14 +1,24 @@
...
 
+RUN mkdir -p /usr/app/ \
+        && mkdir -p /usr/app/frontend \
+        && mkdir -p /usr/app/media \
+        && mkdir -p /usr/scripts \
+        && mkdir -p /usr/shared
+
+WORKDIR /usr/app
+
+RUN git clone https://github.com/AEGEE/oms-applications-frontend.git frontend
```
blacksph3re, line 17: This should not happen in a dev dockerfile. I want to manage my repos outside of docker

docker/omscore/Dockerfile
```diff
@@ -55,42 +43,27 @@ ARG TZ=UTC
...
 # Copy own scripts
 #####################################
-# Done via volume mounts
-#COPY ./.env /root/.env
-#COPY ./bootstrap.sh /root/bootstrap.sh
+COPY ./.env /var/shared/core/.env
+COPY ./bootstrap.sh /var/shared/core/bootstrap.sh
+
+USER root
+RUN git clone https://github.com/AEGEE/oms-neo-core.git /var/www
```
blacksph3re, line 64: Should not happen in a dev environment container, only in production

docker/omsevents-frontend/Dockerfile
```diff
@@ -1,15 +1,23 @@
...
+    
+RUN mkdir -p /usr/app/ \
+        && mkdir -p /usr/app/frontend \
+        && mkdir -p /usr/app/media \
+        && mkdir -p /usr/scripts \
+        && mkdir -p /usr/shared
+        
+WORKDIR /usr/app
+
+RUN git clone https://github.com/AEGEE/oms-events-frontend.git frontend
```
blacksph3re, line 17: Should not happen in a dev environment container, only in production

blacksph3re (commented):
I would not merge it yet as it is not suitable for development

PR #17, opened 2017-06-06
Make it just start

linuxbandit:
Decoupled docker-compose and docker-stack (former is v2 and latter v3)
Dockerfile was corrected
Makefile was removed a command just for safety in case of non-vagrant
use

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -18,8 +18,8 @@ restart: stop bootstrap
 nuke:  
 	sudo docker-compose down -v
   
-clean:
-	sudo docker rmi $(docker images -qf "dangling=true")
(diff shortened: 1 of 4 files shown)


blacksph3re:
You use absolute paths in the dockerfile. Please do not do so, I don't have these paths and a production environment might also not. Either just use named volumes without a path or use relative ones.

linuxbandit:
where?

linuxbandit:
you mean docker-compose, not dockerfile?

blacksph3re:
oh yeah, true, in the docker-compose.yml. I couldn't leave inline comments for some reason

PR #18, opened 2017-06-07
Revert "Make it just start"

blacksph3re:
Reverts AEGEE/oms-docker#17

Reason: dev environment requires absolute paths and doesn't allow having repos outside of containers


linuxbandit:
As agreed, it is better to split the files

PR #19, opened 2017-06-14
Remove named volume mounts

linuxbandit:


i3anaan:
Could you give a bit more clarification on the thought process of this PR, the motivation? Why would we do this? I mean it looks good, but really I cannot tell which way is the better way.

linuxbandit:
The reason is that until now I did everything by the book and used named volumes, and for development reasons I also mounted such named volumes to filesystem. The mount though only happens with absolute paths, and such absolute path is only working out-of-the-box within Vagrant. 

Ideally to keep everything close to a production system we still should use named volumes, and the non-vagrant users should modify the paths at the end of the `docker-composer` file. This way, to make it production ready one can only remove the mount in `driver_opts` and uncomment `driver=local` line.

This change only removes the use of named volumes, to use relative paths.

What are your thoughts?

i3anaan:
I believe this is fixed on `dev` by now, can you verify @linuxbandit ?

PR #20, opened 2017-06-16
Stabilize the dev environment again

blacksph3re:
The recent pushes to create a production environment made development harder, which is why I created another folder with a dev environment (and also a staging one which is temporary until the production is up) 

However I wasn't able to test on windows, it would be good if someone could look into that still


linuxbandit:
Ok, I compared it with [this file](https://github.com/linuxbandit/oms-docker/blob/bbcc8303a06cfca45cb4247a25c50d7b6b5edef0/docker/docker-compose.yml) and that's what I'll refer to when I comment line by line

docker-dev/docker-compose.yml
```diff
@@ -0,0 +1,203 @@
...
+
+    traefik:
+        image: aegee/traefik:dev
+        ports:
+            - "80:80"
+            - "8080:8080"
+            - "443:443"
+        volumes:
+            - /var/run/docker.sock:/var/run/docker.sock
+
```
linuxbandit, line 16: no traefik.toml?

docker-dev/docker-compose.yml
```diff
@@ -0,0 +1,203 @@
...
+        image: tianon/true
+        volumes:
+            - "../oms-core:/var/www"
+            - shared:/var/shared
+
+### Workspace Utilities Container ###########################
+
+    omscore:
+        image: aegee/omscore:dev
+        volumes_from:
```
linuxbandit: `volumes_from` [will be deprecated soon ](https://docs.docker.com/compose/compose-file/compose-versioning/), so it's advisable (yet, silly..) to copy and paste (that's why I repeated them)

docker-dev/docker-compose.yml
```diff
@@ -0,0 +1,203 @@
...
+            - "7000:7000"
+        volumes:
+            - ./docker-compose.yml:/usr/src/docker-compose.yml:ro
+            - ../oms-serviceregistry:/usr/src/app
+            - shared:/usr/src/shared
+
+### OMS CORE        #######################################
+### Applications Code Container #############################
+
+    omscore-data:
```
linuxbandit, line 31: why this instead of a named volume? "data container" pattern was used before the introduction of the named volumes, so now the best practice is to use that

docker-dev/docker-compose.yml
linuxbandit, line 41: `volumes_from` [will be deprecated soon ](https://docs.docker.com/compose/compose-file/compose-versioning/), so it's advisable (yet, silly..) to copy and paste (that's why I repeated them)

docker-dev/docker-compose.yml
```diff
@@ -0,0 +1,203 @@
...
+        expose:
+            - "9000"
+
+### Nginx Server Container ##################################
+
+    omscore-nginx:
+        image: aegee/omscore-nginx:dev
+        volumes_from:
+            - omscore-data
+        links:
```
linuxbandit, line 60: `links` [are already deprecated](https://docs.docker.com/compose/compose-file/compose-versioning/#version-2), the docker way is to use `networks` (but for a dev environment we don't need to differentiate, that's why [I declared them but not used](https://github.com/linuxbandit/oms-docker/blob/bbcc8303a06cfca45cb4247a25c50d7b6b5edef0/docker/docker-compose.yml#L290))

docker-dev/docker-compose.yml
```diff
@@ -0,0 +1,203 @@
...
+            - "traefik.frontend.priority=100"
+            - "traefik.port=8086"
+            - "traefik.backend=omsapplications-frontend"
+            - "traefik.enable=true"
+            - "registry.modules=/getModules.json"
+
+
+### Volumes Setup ###########################################
+
+volumes:
```
linuxbandit, line 195: why no `logs` volume?

docker-dev/omscore/Dockerfile.dev
```diff
@@ -0,0 +1,98 @@
+FROM phusion/baseimage:latest
```
linuxbandit, line 1: why changing from laradock?

docker-dev/omscore/Dockerfile.dev
```diff
@@ -0,0 +1,98 @@
...
+ARG PGID=1000
+RUN groupadd -g $PGID laradock && \
+    useradd -u $PUID -g laradock -m laradock
+
+# Source the bash
+RUN . ~/.bashrc
+
+
+#####################################
+# omscore bootstrap
```
linuxbandit, line 84: I like it if it works from here!

docker-dev/omsserviceregistry/Dockerfile.dev
```diff
@@ -0,0 +1,9 @@
+FROM node:8
+
+RUN npm install -g supervisor bunyan
+
+RUN mkdir -p /usr/src/shared 
+
+WORKDIR /usr/src/app
```
linuxbandit, line 7: One thing I had in mind when I created the dockerfiles was to put everything in the same folder (it wasn't prioritary though, so I didn't). What do you think about this concept of having all the source code in the same folder, no matter of the container?

docker-dev/omsserviceregistry/Dockerfile.stage
```diff
@@ -0,0 +1,13 @@
+FROM node:8
```
linuxbandit, line 1: isn't 8 too bleeding edge?

docker-prod/php-fpm/laravel.pool.conf
```diff
@@ -0,0 +1,76 @@
+; Unix user/group of processes
```
linuxbandit, line 1: In the refactoring work I had removed it because it wasn't used (along with many things), is it used now?

docker-dev/docker-compose.yml
```diff
@@ -64,7 +64,7 @@ services:
         labels:
             - "traefik.backend=omscore"
             - "traefik.port=80"
-            - "traefik.frontend.rule=HostRegexp:{domain:[a-z]+}"
+            - "traefik.frontend.rule=HostRegexp:{domain:[a-z0-9.]+}"
```
linuxbandit: don't you need the `:` too?

linuxbandit (commented):
Also for greater clarity between versions, you could have kept the folder name the same but it's okay, I managed to see the diffs anyway 

wtf this comment also changed all the timestamps of the line-by-line comments? whatever, dates are not important

docker-dev/omsserviceregistry/Dockerfile.stage
serge1peshcoff: @linuxbandit It's pretty stable AFAIK and also there are some benefits instead of using v7 (speed improvements of both node and npm, async/await, util.promisify etc.), so I think it would be better with Node v8. That's what I was going to do with oms-events actually, because I want to use some new fancy features like async/await, it makes code way more readable.

docker-dev/docker-compose.yml
linuxbandit: aha just saw the slack comment, I thought this was to add some port thingy (for which you'd have needed `:`)

i3anaan: I am actually not sure if port notation will be needed when using a different port from port 80, My guess is not as ports are already specified elsewhere in the docker.yml file. As a sidenote: as far as OMS is concerned, we are connecting to the system using the normal port 80 (VM in VM in VM stuff)

i3anaan:
Encapsulated in https://github.com/AEGEE/oms-docker/pull/22

PR #23, opened 2017-07-13
Update master (current default branch)

i3anaan:


blacksph3re:
I am strongly against merging this PR because it would make it really difficult for me to work on oms applications. It's true that I didn't work on it the recent days but still I hope to be able to take it up again. Also, why do you consider the serviceregistry as unused? My thoughts on the staging version I have expressed in the #operations channel and thus I would also be sad if my effort in this is disregarded, though I know that it is not yet a production ready setup (the volume mounts in the docker-compose.yml would have to be adjusted)

i3anaan:
@blacksph3re First off, please note that this version is the latest on the dev branch, this was merged ~4 days ago after me, Fabrizio and Sergey agreed to https://github.com/AEGEE/oms-docker/pull/22

To address some of your concerns:

> Also, why do you consider the serviceregistry as unused?

I removed it and the system still worked, did not notice any degraded functionality.

> My thoughts on the staging version I have expressed in the #operations channel

Taking from my comment on https://github.com/AEGEE/oms-docker/pull/22:

```
Removed differentiation between dev/stage/prod. (containers still have tags).
The reason for this was that the staging files we had were very similar to the dev files we had, causing confusion and danger of forgetting to maintain both files.
In the end, I (and @linuxbandit (who is here)) decided that the need for clarity and easy to maintain code was higher than that of having separate deployment modes.
In the future, we might end up adding it back again, but for now, it is overkill.
```
> and thus I would also be sad if my effort in this is disregarded
```
PS: If anyone can implement this without the need for having extra files, I would be very interested in this! (using an extra parameter + if statements or something along those lines.)
```

> though I know that it is not yet a production ready setup (the volume mounts in the docker-compose.yml would have to be adjusted)

**It was not just not production ready, it was also not development ready. We needed stabilization, we needed moving dockerfiles to the services' repositories and we needed a check if the integration of the entire system worked.**


Some extra comment:
Please do not regard this as final and perfect, but for the state of development that we are in, this was necessary. I still very much welcome some of the changes you proposed, the difference between dev/stage/prod is cool, yet more of a future thing. Additionally, I know there are some things wrong regarding usage of docker, but once again: This was better than what we had. Regarding pulling in Master: I figured why not? This is probably the most stable it will be in a while.

linuxbandit (approved):

linuxbandit:
Only thing I do not understand is looking at the git history: my changes are there and also nico's after mine (i.e. `traefik` is there and `stabilise` too - but they were 2 parallel ones!)

Other than that it is good as discussed

README.md
```diff
@@ -1,32 +1,39 @@
...
-cd oms-docker/docker-dev
+cd oms-docker/docker
 sudo docker-compose up
 ```
 
-NOTE: Because Traefik works reading the URL of the request, it is necessary to add to your `/etc/hosts` file a name for your address (127.0.0.1 if bare Docker, 192.168.192.168 if Vagrant) - Traefik is configured under the name **appserver**.
-
-Open you browser on `http://appserver` and login with the default credentials. Navigate to the modules tab and activate all the microservices
-
+*This is assuming you are running a Linux installation, for other operating systems virtualization is recommended.*
```
linuxbandit: Where is it written about the virtualisation (how to virtualise)?

linuxbandit (approved):
Tag it :p 

0.4.0 ?

README.md
i3anaan: The more detailed installation instructions contain a link to an article about it:
https://oms-project.atlassian.net/wiki/spaces/GENERAL/pages/17235970/Installation

If you still want to do vagrant stuff feel free, but to be honest I felt like this was the simpler solution: giving the User more freedom in terms of how to get Ubuntu running, be at native, VirtualBox, Vagrant, or any other way. Besides, setting up VirtualBox is not really hard and should be well within the capabilities of someone supposed to work on the project.

linuxbandit: Yes I wanted to ask if it was there (I am limited on a small screen so bear with me for not checking)

The vagrant is more for a 'one-liner' that makes it set up, that's why I asked

i3anaan:
NOTE: Default branch has been changed to dev.

i3anaan:
Let's close this until we have `oms-events` and the `registry` in `dev` again

PR #26, opened 2017-07-24
Core env files update.

i3anaan:
With a sligth GIT mess / confusion.


linuxbandit (approved):
seems ok, why removing the seeder (and why was it there in first place?)

i3anaan:
The seeder was being overwritten because it needed to change the admin password on live servers. However instead of overwriting the entire file, this commits moves the admin password to the .env file which needs to be overwritten anyway.

PR #27, opened 2017-07-25
Corrected TAG > RESET in oms_setup.sh

i3anaan:


linuxbandit (commented):
what does it do?

i3anaan:
When the oms_setup.sh script prints the configured parameters it printed "TAG" which was actually a copy paste error and should have been "RESET"

The line changed actually explains it pretty well

linuxbandit (approved):
hahaha sorry I was absolutely confused by this because in my mind it was about the git tag, for some weird reason I only looked at the red part

PR #28, opened 2017-07-25
Improved testing: Added local loopback test.

i3anaan:


i3anaan:
FYI: This also includes the latest oms-core commit.

linuxbandit (approved):
Nice! Are you planning to add more ? Should be fine to find something that lets API testing

i3anaan:
@linuxbandit The idea was that this would only test the network configuration/setup (mainly travis / omscore-nginx). For this I felt a simple ping would be enough, leaving functional API testing to the core tests. (This does however currently act differently due to the php-fpm setup https://oms-project.atlassian.net/browse/CORE-112).

As the system increases in microservices it could be cool to have some combined actions tested, for now this is not the case though.

i3anaan:
NOTE: Default branch has been changed to `dev`.

PR #29, opened 2017-07-28
Update vagrantfile to current status of docker composer

linuxbandit:
The vagrant vm is set to the new status of docker-compose file (no folders for services which aren't there)

to test,
 1. `git clone --recursive <repo>`
 2. `cd oms-docker`
 3. `vagrant up`


i3anaan:
Crap, I guess I should have reviewed this earlier, didnt really realize that. @linuxbandit Could you check if this is still somewhat up-to-date?

linuxbandit:
Don't worry, it is ok!

It is basically always up-to-date because this is a simple abstraction, if the docker works then the vagrant automatically works... I will only have to change the bootstrap command in Makefile (oms.sh? idk the name anymore), I will push some changes that solves the conflicts

PR #30, opened 2017-08-12
Docker-compose split

blacksph3re:
I split the docker-compose files into small ones that live inside each repository, to continue on the path of the previous docker split. All docker-compose partials will be merged together before execution by the docker-compose.sh script, so the script should be used instead of the docker-compose command, like:

./docker-compose.sh up -d

Before merging, please merge all of these:
https://github.com/AEGEE/oms-neo-core/pull/47
https://github.com/AEGEE/oms-events/pull/22
https://github.com/AEGEE/oms-events-frontend/pull/7


blacksph3re:
The PR test somehow stalls at composer install while the push test succeeds. I don't know why, did someone experience something similar? @i3anaan ?

i3anaan:
@blacksph3re The difference between the PR and push tests is that the PR merges it into the latest default branch (dev) if I am not mistaken. What might be happening is some kind of merge conflict. Try merging dev into this branch again and see if that opens up any issues, or at least lets you reproduce the issue that Travis is running in to.

i3anaan:
Welp, reran the test and that seems to have fixed it  ¯\\_(ツ)_/¯

i3anaan:
I noticed your PR on oms-docker so went ahead and looked at it. I am amazed by the similarity between our work (https://github.com/AEGEE/oms-docker/tree/docker-upg for mine), I guess we at least have the same idea? :stuck_out_tongue:

Few things to note:
- What I understand from GitHub is that yours 'compiles' the docker-compose files in a new, single, docker-compose file?
- - If so: I like my approach of keeping the files independent but connecting to the same network more.
- You readded some containers/services, do these have automated tests?
- - Personally, I'd say the oms-events/oms-application is spewing too much in the docker log right now.
- You seem to be better with bash scripts than I am.

blacksph3re:
I am happy the tests now pass for some magic reason. To comment your notes

1) Yes exactly.
2) Ehm events is on the verge of getting them, the registry doesn't... And yes I know that they spam the logs quite much, if you want I can fix that for the registry, but it still has this hack to register everything to the core instead of the other way around, so I guess it will need a code-cleanup anyways, as soon as the core fetches the modules from the registry. About events, tbh I haven't done much on it in the past, @serge1peshcoff  is the one to address
3) Idk, I was pretty impressed by your oms-build.sh ^^

blacksph3re:
And something with the tests is strangely nondeterministic. Now the one fails for a segfault, after I have changed something completely unrelated ^^

serge1peshcoff:
I've actually fixed crashing and log spamming on oms-events here https://github.com/AEGEE/oms-docker/pull/21
Not sure about the oms-applications though, since I didn't look into them.

blacksph3re:
@serge1peshcoff ups, I didn't see it. I also fixed crashing and log spamming in https://github.com/AEGEE/oms-events/pull/22, though there are still some bugs somewhere in the code =)
I left oms-applications out of this commit as I personally will probably not have time to dig into it soon, so I leave this for the one who is willing to take up this challenge.

i3anaan:
I took a closer look at this and combined, in my opinion, the best of both in #31

Some things I noticed about this version:
- oms_setup.sh is broken yet not removed (including the split-out build.sh files??)
- I think the non-deterministic issues your tests run into might be caused by the way you are combining the docker commands: #31 solves this by specifying networks and launching them one after another.
- Testing is not streamlined
- I do not like calling it docker-compose.sh, as it honestly does not really have anything to do with it, all it does is combine several services into 1 system, which we call OMS (hence I named it oms.sh in #31 )
- I think having oms-globals is not necessary, all the things you have placed in there belong to oms-docker in my opinion.

I believe #31 improves on all of these issues I have with this while still having the same overall splitting approach.

I realize #31 did not integrate events/serviceregistry yet, if you approve on what it is for docker and for core I will add this.

i3anaan (changes requested):
See earlier comment, I favour #31 over this PR.

linuxbandit (changes requested):
I agree with Derk about the keeping the files independent, and using the network feature. It is anyway the good practice. 
Also about the naming, ons-globals is not necessary as well as the docker-compose.sh not being related to it

Also personally, I prefer the “superscript” of Derk that can decide when to build and when to push to the dockerhub (not in this PR). It can let the operator decide when to perform maintenance on the images, and it has the commands that a bootstrap needs (doing them all at once could raise the build time, which sucks during dev)

i3anaan:
Since #31 was merged, this branch is now obsolete

PR #31, opened 2017-08-16
Docker upgrade

i3anaan:
Splits docker-compose.yml accross repostiories, efftively creating services with microservices.
Upgrades some of the docker-compose.yml files:
- Upgraded version where possible
- Added build information which allows commands such as `docker-compose build`
- Specifies 'OMS' docker network.
Greatly streamlined automated testing instructions
Replaced oms_setup.sh with oms.sh, which I consider to be better in just about every way (also the oms_setup.sh does no longer work with the new docker-compose.yml split)

See https://github.com/AEGEE/oms-neo-core/pull/49
Very similar to https://github.com/AEGEE/oms-docker/pull/30


i3anaan:
I Consider this to have the following advantages over https://github.com/AEGEE/oms-docker/pull/30

- Slightly improved docker-compose.yml files.
- Includes testing + deployment streamlining
- Specifies the docker network
- Offers some more flexibility in the 'main' script: `oms.sh` by providing build/down/push options.
- Does not have deprecated files such as docker/oms_setup.sh and oms-core/docker/build.sh

i3anaan:
Left a detailed comment on #30, worth reading as well when considering this PR.

Important part: I will add support for events / registry if you approve of the approach I took for docker/core

Additionally, I am open for some naming suggestions/discussions if that proves to be the point of disagreement.

blacksph3re:
I agree with adding networks, it's nice to isolate stuff from each other, but I won't approve this PR before the service registry and one ms is in here as I have 2 doubts it works and one improvement suggestion for the oms.sh script. 

Doubt 1) does volume sharing work in your setup? 
Doubt 2) does the network thing introduce any kind of mutual unreachable containers 
Improvement 3) instead of calling docker-compose up in every folder with some predefined commands it might be nice to just forward the command line arguments as I did in my script. This way you don't lose support for something like docker-compose logs containername or any other docker features that might be nice to use at some point. Instead of cd-ing into the directories I use the - f option, maybe you can have a look at this

i3anaan:
Good to hear you like the network!
I will look into adding the serviceregistry and events back in then.

To answer your questions:
1) No, if I understand correctly we already were using a network before (in fact `docker-compose` pretty much does `docker` commands and combines it within a network - everything done with `docker-compose` can be done with `docker`). So pretty much what changed is that we explicitly set the network to connect to, making it easier/better.
2) Same as above: nothing should be changed.
3) Will have a look at this!

I will also read up a bit more about the networks things to check I am understanding it correctly.
Will probably be tomorrow/later

blacksph3re:
Okay, in case you opt for the -f option, my doubt in 1) are gone as my attempt worked with the -f option. I know that docker-compose just uses docker behind, but it renames some things, like having containernames like omsdocker_omscore_1 and also other volume names. So I can imagine that in case docker-compose already finds another volume with the same name from a previous invocation, it might rename it to omsdocker_shared_2 or something and it ends up being different things. So yeah, it'd be nice if you either check that out or go for the -f option (which would also bring the advantage of 3) ). But notice that in the -f option file-paths are relative to the first docker-compose.yml, which is why I passed an empty docker-compose.yml in the root folder at first.

i3anaan:
Found this page:
https://docs.docker.com/compose/extends/

After reading it I fully agree with you to use the -f option.

i3anaan:
A couple of new things with the latest 2 commits:

- Upgraded docker-compose version to 3.3
  - This removed volumes_from, making omscore_bootstrap obsolete
- To keep individual service clusters (such as omscore + omscore-nginx + etc) indivually runnable by `docker-compose`, I added a .env file that is used by docker-compose when executing through `oms.sh`. When running such a cluster individually the missing variables default to empty, still resulting in a correct docker-compose.yml
- If we want to support dev/stage/prod environments again we should add a command to `oms.sh` that makes it load extra docker-compose.yml files that override the default configuration.

oms.sh
```diff
@@ -1,75 +1,25 @@
...
 
-echo -e "OMS: Create network..."
-docker network create OMS
-
 ## Declare docker-compose.yml folders
-declare -a arr=("docker" "oms-core/docker")
-
-## Save base path
-BASE_PATH=$(pwd)
+declare -a services=("oms-global" "oms-core")
```
blacksph3re: What about instead of declaring existing services just go through all subfolders and if it has a docker-compose.yml it will be included? This way there is 1 step less in adding a new service.

i3anaan: I disliked this because of 2 reasons:
- I like actually having to specify which services to include (makes it easier to disable/enable a single service)
- Secondly, it allows specifying an order to the docker-compose files, something which might not be relevant yet but could be really useful later on.

blacksph3re:
I am not sure about the idea with the .env files for isolated running or running in the whole system... Currently the configuration works differently in the microservices, some having a config.json, .env, or something completely different. We could however standartise it in a way, I post on slack for this

i3anaan:
@blacksph3re The .env addition is currently only used to resolve relative paths, this is purely for the docker environment, the services should not be impacted by this (the services do *not* have a .env file, making it work as intended)

Without the .env file present, we either need full path names (doesn't work with multiple developers) or we it would not be possible to run the docker-compose.yml files individually (from their folders)

serge1peshcoff (approved):
I don't have the ability to test if it's actually working but I like the implementation.

.env.example
```diff
@@ -0,0 +1,4 @@
+#Environment file for docker-compose
+
+PATH_OMS_GLOBAL=oms-global/docker/
+PATH_OMS_CORE=oms-core/docker/
```
blacksph3re, line 4: Might it be possible to generate those on the fly?

oms.sh
blacksph3re: Mh, okay, I like doing point 1 through commands to docker-compose (docker-compose up -d omsevents omsevents-frontend), but I guess this is personal preference. With the order, idk if that ever becomes useful because it only affects how the docker-compose file is merged, not in which order they are started as this happens in parallel. And I don't think we ever get to a point where the merging of these files needs to happen in an order, but I am not sure either.

blacksph3re (approved):
Except for some minor in-line comments I like the changes - haven't tested them though but I believe in travis ^^

oms.sh
i3anaan: The order can become relevant/be used for specifying different environments! (dev/stage/prod)

.env.example
i3anaan: Would that really be worth it? I imagine we might very soon add other environment variables to this file, I do not really see harm specifying this here.

i3anaan: I suggest looking into this combined with the general config/environment rework you were thinking of.

PR #32, opened 2017-08-18
Updated php-fpm dependencies to include php7.1-intl

i3anaan:
Required by core to transliterate 'weird' characters:
https://github.com/AEGEE/oms-neo-core/blob/develop/app/Facades/Util.php

Also see:
https://github.com/laradock/php-fpm/issues/13


i3anaan:
While at first thinking there was something wrong with Laradock/php-fpm, this has by now been resolved: there was nothing wrong with it, I used it incorrectly.

blacksph3re (approved):
I approve!

PR #33, opened 2017-08-28
Re-intergrate registry and events

i3anaan:
(This is pretty much a renamed branch of `dev-registry`, ignore that one and use this one)

Re-integrates `oms-events`, `oms-events-frontend` and `omsserviceregistry`

Merge order:
~1. https://github.com/AEGEE/oms-serviceregistry/pull/2~
~2. https://github.com/AEGEE/oms-events/pull/23~
~3. https://github.com/AEGEE/oms-events-frontend/pull/9~
4. This PR

Im not saying everything is ready to merge just yet, but this provides a clear set of branches/PR to work on as well as an order.


i3anaan:
@serge1peshcoff Can you add the following tests to the docker travis.yml:
- The tests that are run for checking oms-events (not sure how it worked from that travis.yml)
- Expand the localloopbacktest to include a ping to oms-events
- Maybe some simple check to see if omsevents can talk with omscore?

i3anaan:
Then some log things that should change:
- omscore should silence the migration
- omsevents should silence npm
- omsserviceregistry should try to bootstrap less frequently and not print the entire loaded docker file (either write it out, or dont do it at all)
- postgres and mongodb should be silent when working (right now there are all kind of warning and messages and stuff that imo can be silenced)

I think this would go a very long way.
We can also use an environment variable to specify log level? I think this is almost required by now, we can look into doing this right away maybe?

i3anaan:
@blacksph3re Can you maybe implement a simple test to see if the serviceregistry is working? Or would checking for the service running be enough?

blacksph3re:
Yes I will rewrite the serviceregistry a bit, the code is pretty messy. I hacked it together as I didn't know back then which role it would eventually take, but I think things are a lot clearer by now. =)

blacksph3re:
Okay, the serviceregisry is rewritten and less verbose. And are you actually looking through the postgres and mongodb logs? I never found anything useful there so I just leave them out of my consideration.

Also fully silencing logs is something I would actually not do for production, as there it is always good to have as much information as possible. When something crashes, even in the unlikely case that it's the db, it would be good to have a prior log so we know what lead to the crash and might be able to debug it. For dev it's okay, but in case we ever roll this out we should remember to unsilence the logs.

blacksph3re:
Also my alastair PR goes into this branch, so we should probably close https://github.com/AEGEE/oms-docker/pull/35 first, and it would appear here as well

i3anaan:
@blacksph3re Agree with the log silencing, which is why I would have liked to be able to set log level, but that doesn't seem possible right now.

Yes, I have it on my list to look into the DB spams.

About Alastair: I am not sure, I mean if everything has tests its okay I guess, otherwise it is just more risk of things breaking.

blacksph3re:
I think Alastair is currently the most thoroughly tested piece of software in the stack with 89 passing tests :) I worked partially test-driven.

linuxbandit:
some considerations about this PR (I am trying to figure the general state out)

- if this is basically `dev-registry` what is the difference? we could just rename branches, for the future 
- for the future, it is better to have meaningful names for the branches
- for the future it is also better to have not too many branches (and this is why one does not work on upstream but on their own fork, and only make a PR out of meaningful things - so all the rubbish branches stays on the fork)
- what is keeping this to be merged? only the other ones?
- why is there a "dependency" on other repos? ("When they will be merged, this PR will be updated with the latest commits of those repos" - is that my answer?)

Thanks for helping me out, I am getting back to the development now that agora craziness is over :)

linuxbandit (approved):

linuxbandit:
I am still not happy about the random names, but as we are at the hackathon being productive it's good to celebrate :p

PR #34, opened 2017-08-30
Changed enabled services to be parsed from .env

blacksph3re:


i3anaan:
These tests seemed to have failed because of Travis ?? reasons, I am rerunning them

i3anaan:
Found a bug: Tests were checking if things such as `oms-events` where up, however in your `.env.example` you did not enable them by default. I changed it to enable all services by default (removed alastair as that is not yet in)

PR #41, opened 2017-10-07
Update vagrant environment

linuxbandit:
This is (yet) another update (I will close #29), in which the VM for developers of windows and mac is updated with several features:

- add updated OS to reflect what we'll have on fraktis
- add updated bootstrap method
- add utility containers for orchestration and CI 

README was updated accordingly


i3anaan:
Looks really good.
What is portainer?

i3anaan (approved):
Under the assumption that you have tested the vagrant installation I approve of this!

linuxbandit:
You can test it yourself too ;) and you can get a hang of what is portainer (I just briefly showed you at agora, it's basically an orchestrator but simpler than rancher)

`git clone --recursive -b update-vagrant https://github.com/linuxbandit/oms-docker`
`cd oms-docker`
`vagrant up`

note that for that you *must* edit your `/etc/hosts` because traefik serves it through "portainer.*" e.g. http://portainer.appserver/ only (I mean, also with http://portainer.localhost:8888/ but you should edit the hosts anyway so why not using `appserver` :p )

linuxbandit:
will merge tomorrow after andrea tries it

PR #43, opened 2017-12-15
 More error resistant oms-core bootstrap

i3anaan:
Should squash and merge because I messed up some git things


i3anaan:
Already included

PR #44, opened 2017-12-17
Print OMS logo on start up!

i3anaan:


linuxbandit (approved):
no bananas were spotted

PR #48, opened 2017-12-23
Pgadmin - fixed

i3anaan:
Wasnt working before, now it is, also set the credentials to admin:admin.

Still not reachable through pgadmin.appserver/, not sure why.


i3anaan:
@linuxbandit Do you by now understand how to fix it so that it is reachable at `pgadmin.appserver/`?

oms-global/docker/docker-compose.yml
```diff
@@ -50,15 +50,18 @@ services:
     pgadmin:
         image: fenglc/pgadmin4:2.0
         restart: on-failure
-        expose:
-            - 5050
+        ports:
+            - 10100:5050
```
linuxbandit, line 54: let me crosscheck with what i have here because I am quite sure it is not needed to forward the port

i3anaan: Forwarding the port allowed me to access it on appserver:10100 though, might be a nice additional feature to have (if for whatever reason you cant (or forget) to update your hosts file)

linuxbandit (approved):

PR #50, opened 2018-01-23
[wip] improving a bit of this and that

linuxbandit:


linuxbandit:
there's nothing to review now, I just had a bug in the cli and could not see which files are conflicting

oms.sh
```diff
@@ -22,7 +22,7 @@ service_string=$(printenv ENABLED_SERVICES)
 ## Split services into array
 services=(${service_string//:/ })
 
-command="docker-compose -f base-docker-compose.yml"
+command="sudo docker-compose -f base-docker-compose.yml"
```
serge1peshcoff, line 25: Just for information, you can run Docker for non-sudo users: https://askubuntu.com/questions/477551/how-can-i-use-docker-without-sudo
Not sure what be more secure though.

linuxbandit:
this is closed as PR because it works better as local branch of upstream. Also I will improve the commit message which was just a fast wip

oms.sh
linuxbandit: thanks! I will think about it in the optics of continuous deployment, because right now I am just thinking development

PR #51, opened 2018-01-25
Final improvements before the merge to dev

linuxbandit:
Last fixes before this is stable enough to enter dev

Changing the name of the branch for the merge accordingly

(more to add, if spotted)

PR #52, opened 2018-01-25
Add script for automatic deployment on cron job

linuxbandit:
While in the future an approach hook-based will work better (since
state-of-the-art) for now this will do. The deploy script is meant to be
run with privileges (to access the docker daemon) and to be scheduled as
a cron job: there will be a check if updates of upstream and a pull
with update of submodules, and a restart of the services (docker-compose
will take care of restarting only the changed containers).
There are also other options to 'refresh' the installation by deleting
the .env file, or to nuke it and re-downloading it from scratch.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -11,4 +11,6 @@ PATH_OMS_LOGO_PRINT=oms-logo-print/docker/
 
 
 # To enable services, add them to the array, like
-ENABLED_SERVICES=oms-global:oms-core:oms-serviceregistry:oms-events-frontend:oms-events:alastair:oms-logo-print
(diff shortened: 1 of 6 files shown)


linuxbandit:
This should only be merged after the last polishment of #51 (e.g. the name of the branch changes to `dev`)

linuxbandit:
Just in case, the change was `git clone --recursive --branch dev https://github.com/AEGEE/oms-docker.git 1>/dev/null 2>&1`
(dev instead of feat-autodeploy)

serge1peshcoff (approved):

PR #53, opened 2018-02-06
Basic absolute minimum of oms-static

i3anaan:
Really more of a hotfix than anything else, will deploy oms-static and oms-legacy on fraktis, which is why I am PRing this.


linuxbandit:
Me and Sergey are cleaning up the branches

PR #54, opened 2018-04-30
New core stable

i3anaan:
Fixes fresh launch issues, minmizes default services


serge1peshcoff (commented):
Can we leave oms-events and oms-events-frontend there for a while?

PR #55, opened 2018-07-05
Mvp last tweaks (cleanup of structure)

linuxbandit:
In theory this is the working backend without frontend yet. I am going to test it now, only the api should be successful anyway. Will add a reviewer when I finish testing


.env.example
```diff
@@ -1,14 +1,14 @@
...
 
 COMPOSE_PROJECT_NAME=myaegee
 PATH_OMS_GLOBAL=oms-global/docker/
-PATH_OMS_CORE=oms-core/docker/
 PATH_OMS_REGISTRY=oms-serviceregistry/docker/
 PATH_OMS_EVENTS=oms-events/docker/
-PATH_OMS_EVENTS_FRONTEND=oms-events-frontend/docker/
-PATH_ALASTAIR=alastair/docker/
+PATH_OMS_CORE_ELIXIR=oms-core-elixir/docker/
+PATH_OMS_LOGINSERVICE=oms-loginservice/docker/
```
serge1peshcoff: do we need login service? since it's not used and it's in the core now

.gitmodules
```diff
@@ -1,21 +1,18 @@
...
 	path = oms-logo-print
-    url = https://github.com/AEGEE/oms-logo-print.git
+  url = https://github.com/AEGEE/oms-logo-print.git
 [submodule "oms-global/docker/portal/files"]
 	path = oms-global/docker/portal/files
 	url = https://github.com/AEGEE/AEGEEPortal
+[submodule "oms-core-elixir"]
+	path = oms-core-elixir
+	url = https://github.com/AEGEE/oms-core-elixir.git
+[submodule "oms-loginservice"]
```
serge1peshcoff: and here as well

.env.example
```diff
@@ -1,14 +1,14 @@
...
 PATH_OMS_LOGO_PRINT=oms-logo-print/docker/
 
 
 # To enable services, add them to the array, like
-ENABLED_SERVICES=oms-global:oms-core:oms-serviceregistry:oms-events-frontend:oms-events:alastair:oms-logo-print
+ENABLED_SERVICES=oms-global:oms-core:oms-serviceregistry:oms-events:oms-logo-print
```
serge1peshcoff: did you mean `oms-core-elixir` instead of `oms-core`? :D

serge1peshcoff (changes requested):

.env.example
linuxbandit: i included it just to be sure but I agree with you

.env.example
linuxbandit: thanks! I will correct it

serge1peshcoff (approved):

PR #56, opened 2018-08-12
Add https auto configuration with LetsEncrypt

linuxbandit:


blacksph3re:
Ah brilliant... on the first looks it seems like it would work, I didn't remember you could explicitly specify domains

linuxbandit:
tried with another domain and it works nice, so I put the "target domain" to be ready. Later on we can add more subdomains (but it'll be also up to Wim for adding them to dns)

I talk about subdomains because it's much easier to configure traefik for a subdomain than not for a HostPath (and maybe hostpath is going to be modified by our app? we should avoid overwriting)

blacksph3re (approved):
I will merge it as it seems to work

serge1peshcoff (approved):
I trust you on that :D

PR #64, opened 2018-11-16
Bump submodules

linuxbandit:
Because why not

Actually, let's have a discussion and create a nice "strategy" to bump at certain times. See MEMB-344


serge1peshcoff (approved):

linuxbandit:
I will merge it later. #65 needs a newer version of `oms-core-elixir`

MEMB-344 still to be discussed btw

PR #65, opened 2018-11-20
Improve bootstrap script

linuxbandit:
the ideal case is to use Make to run the target you need - no more hassle

The only limitation is when logs are called: if one wants to make logs only for a specific container, they have to use directly `helper.sh --logs containername1 containername2 ...`

Vagrant makes use of that makefile to bootstrap. Vagrant has now 3GB of memory since we have a bigger machine on the cloud.

Makefile also has function to help bumping the submodules at once: assuming all submodules are stable at any given commit, it should be safe to call it every now and then


serge1peshcoff:
can you update README to reflect these changes?

serge1peshcoff:
and probably specify here not only the commands that people need to run, but also some common commands, like executing a process in the container, getting log output etc.

linuxbandit:
Nice one about running the command in the container! Didn't think of it, I will add it

linuxbandit:
Check it out now (easier to see [here](https://github.com/linuxbandit/oms-docker/blob/improve-bootstrap_script/README.md))

README.md
```diff
@@ -10,29 +10,29 @@ It makes use of docker, and docker-compose.
...
 cd oms-docker
 vagrant up 
 ```
+**It will take a while**, so grab a cup of some hot beverage.
 
 [**More information on the installation**](https://oms-project.atlassian.net/wiki/spaces/GENERAL/pages/17235970/Installation)
 
 ## Usage
-After running the system, you can navigate to it in your web browser. The URLs differ based on how you run it; however no matter how you decide to run it, _it is suggested_ to edit the `/etc/hosts` file to add the entry: `192.168.168.168 appserver traefik.appserver portainer.appserver`, to be able to use advanced features.
+After running the system, you can navigate to it in your web browser. The URLs differ based on how you run it; however no matter how you decide to run it, _it is suggested_ to edit the `/etc/hosts` file to add the entry: `192.168.168.168 my.appserver traefik.appserver portainer.appserver`, to be able to use advanced features.
```
serge1peshcoff: 1) I suggest putting other addresses as well (like kibana and my) to hosts
2) reflect here that on Windows it's `C:/Windows/system32/drivers/etc/hosts`

vagrant-post-script/orchestrate_docker.sh
```diff
@@ -1,8 +1,8 @@
...
 
-#cd oms-docker/docker
-#make bootstrap
 
 cd /vagrant/
 #possible to use flag -d
-./oms.sh up -d
\ No newline at end of file
+make bootstrap
+sleep 90 #to give time for the bootstrap
```
serge1peshcoff: I suggest adding proper healthcheck here, something like

```
while (true) {
  if (fetch(oms_core_elixir_address)) {
    break;
  }

  sleep(2);
}`

(no idea how it would be in bash, so it is pseudo JS).
Not critical though, but would be nice, so we'd know if everything has really launched.

File: README.md [hunk 3]
```diff
@@ -42,39 +42,58 @@ For using it, it becomes:
...
 Apache License 2.0, see LICENSE.txt for more information.
 
-## Ports registered on traefik
-| Port | What | Container |
-|---|---|---|
-| 80,443 | standard entrypoint | traefik |
-| 5432 | not registered | postgres |
 
 ## Deployment script
 The types of deployment:
```
serge1peshcoff, line 99: can you add extra \n after this line? so Markdown would understand it's a list and display it properly

serge1peshcoff (changes requested):

README.md
linuxbandit: I don't understand, this part has been removed

vagrant-post-script/orchestrate_docker.sh
linuxbandit: I wanted to avoid potentially infinite loops (if core is broken, e.g.)

Also, the address of frontend should be there, not of core? I guess curl is the real world implementation of what you suggested

README.md
serge1peshcoff: no, the line after The types of deployment

vagrant-post-script/orchestrate_docker.sh
serge1peshcoff: 1) we can add some threshold to exit it, e.g. 5 minutes
2) I'd say the backend. For example, the frontend is loaded but the backend is still loading (was the case for me at least, as somehow building oms-core-elixir takes a lot of time). Then the system will be still unusable till oms-core-elixir will start

README.md
linuxbandit: 1. My is already there, but kibana not. I don't think others (e.g. status, and future ones) will be relevant
2. Yesterday I wanted to get it done for the simple case, though today I see that powershell has some more advanced features! I was thinking that cmd was shit (and it is), but ps is not as bad. I will add a .ps1 file that does the same

serge1peshcoff: 1. sorry, didn't notice. then it's all good

linuxbandit: 2. I wasted all my time until now, fuck windows. I will just include a batch file that must be right-clicked and run as admin, it opens the hosts file in notepad and the user has to paste and save, and invoke vagrant up manually. Fuck this shit

serge1peshcoff: maybe just put 'to edit the `/etc/hosts` (on Windows, it's `C:/Windows/system32/drivers/etc/hosts`)' instead to Readme and let people figure that out by themselves?

vagrant-post-script/orchestrate_docker.sh
linuxbandit: I thought anyway that we should add some Healthcheck in the services (it was in my mind since long), to both know about their health and to accommodate situations like this one (and swarm, or consul, or portainer, or whatever like curl can know how to assess container's health).

Just a simple endpoint like `/ping` or `/health`

serge1peshcoff (approved):

vagrant-post-script/orchestrate_docker.sh
linuxbandit: For now all it does is to check if I got the `<html>` tag, which is the error that Phoenix gives because it is not a known route

linuxbandit: Also, I do that by changing the definition in `docker-compose.yml` of oms-core-elixir from `expose` to `ports`... i.e. I would need a PR on it now

PR #67, opened 2018-11-26
Add basic MTA on a container (MEMB-299)

linuxbandit:
The MTA was supposed to be native, but for a starter we all agreed to
have it in a container. This is by no means complete as it needs testing
by the app etc, but the most basic settings are there

### DIFF
--- a/oms-global/docker/docker-compose.yml
+++ b/oms-global/docker/docker-compose.yml
@@ -32,8 +32,8 @@ services:
         restart: always
         expose:
           - 25
-    #   volumes:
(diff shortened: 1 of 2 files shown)


serge1peshcoff:
I suggest doing it in oms-mailer as it is tightly coupled with it, what do you think?

linuxbandit:
true! Nico you can take the main.cf, or merge, or whatever :)

blacksph3re:
oh I didn't see this one. I will replace my version with this one, I added one to oms-mailer too.

linuxbandit:
then close it when you want

linuxbandit:
@blacksph3re I'd say we can close this, right?

PR #68, opened 2018-11-27
(Impr) introduce basic security concepts around

linuxbandit:
In an attempt to standardise some content (MEMB-180 & MEMB-104) we put
some passwords as single default ones. The rationale is that if we have
a system that can login with 5ecr3t, we forgot to set something -> we
are idiots.

So, it's a 'standard' for the new passwords, in short :p

Also some routes are behind (basic) auth by traefik


linuxbandit:
the thing is that we also have to change the other git submodules, both in docker-compose files and in config.json files (for the node ones, idk the equivalent for Elixir).

Can you guys make a sort of central .env file for every project, such that docker-compose and config.json reads from there?

blacksph3re:
It would be nicer to read from docker secrets for the time when we migrated to swarm, right? Most containers support reading secrets like pw from file.

serge1peshcoff:
@linuxbandit what's the status of this PR? other than merge conflict that can be most likely resolved by rebasing on master?

PR #75, opened 2019-01-29
Fix travis

linuxbandit:


.travis.yml
```diff
@@ -8,17 +8,10 @@ services:
...
-    - bash ./oms.sh exec -T -d postgres-oms-core-elixir true 2>/dev/null 
-    - bash ./oms.sh exec -T -d pgadmin true 2>/dev/null 
-    - bash ./oms.sh exec -T -d oms-core-elixir true 2>/dev/null 
-    - bash ./oms.sh exec -T -d oms-events true 2>/dev/null 
-    - bash ./oms.sh exec -T -d oms-frontend true 2>/dev/null 
-    - bash ./oms.sh exec -T -d mongodb true 2>/dev/null
-    - bash ./oms.sh exec -T -d omsserviceregistry true 2>/dev/null 
-    - bash ./tests/LocalLoopbackTest/LocalLoopbackTest.sh
+    - bash ./helper.sh --execute -- -T -d traefik help 2>/dev/null 
+    - bash ./helper.sh --execute -- -T -d oms-frontend true 2>/dev/null
```
serge1peshcoff, line 17: can you also put oms-events and oms-statutory there?
also keep in mind that oms-frontend is not meant to be run forever, it just builds stuff and exits, so there can be cases where it's not running and the command is executed against it, causing the test to fail

serge1peshcoff (approved):
I approve generally

.travis.yml
linuxbandit: honestly i don't even know if this is a good way of testing them, i translated how it was to how it should be now - and at least a forgotten definition of secrets (e.g., taking the example of the last break that happened), or the lack of a folder (that you fixed) would have been spotted

PR #76, opened 2019-01-30
(bump) Bump version of the submodules via make bump

linuxbandit:


serge1peshcoff (approved):

linuxbandit:
This is so bad, on my test server it fails (because `postgres:latest` is 11, but the files are for pg10)...

We need to version all the containers, `:latest` is not an option

linuxbandit:
pg doesn't start, therefore oms-core-elixir quits because no connection

PR #77, opened 2019-01-30
Do not merge, Test breaking

linuxbandit:
I want to see if with a bigger timeout the problem that arose in PR #76 is spotted


linuxbandit:
this is just because I forgot I also set travis to take from my repo XD 

whatever

linuxbandit:
weird, it quits on oms-frontend (as it should be), but not on oms-core-elixir

PR #80, opened 2019-01-30
(hotfix) Bump version of the submodules via make bump

linuxbandit:
This is for the hotfix where the version of statutory/events' postgres was 10.6 but 11 was required


linuxbandit:
there are more files than should be

PR #82, opened 2019-01-30
Elk rebased

linuxbandit:
supersedes #62 

MEMB-100
MEMB-288

PR #83, opened 2019-02-12
THIS IS REALLY IT

linuxbandit:
XD

The difference is that PR #68 is mandatory here, basically, because kibana also controls the 
   ES instance -> we need traefik to do at least basic auth.

Therefore I close #82 too


linuxbandit:
ok I need to add the submodule still -.-

linuxbandit:
beginning of MEMB-180 & MEMB-104
MEMB-100
MEMB-288
MEMB-289

serge1peshcoff (approved):

PR #84, opened 2019-02-28
Added docker command to helper script

blacksph3re:
This way currently unsupported docker commands can be accessed again through the helper script

### DIFF
--- a/helper.sh
+++ b/helper.sh
@@ -126,6 +126,7 @@ restart=false;
 nuke=false;
 bump=false;
 execute=false;
+docker=false;
 command_num=0;
 declare -a arguments # = EMPTY ARRAY
 if [[ "$#" -ge 1 ]]; then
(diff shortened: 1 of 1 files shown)


linuxbandit:
Can you provide usage examples?

serge1peshcoff:
Actually it can be quite handy when we want to run some custom command. For example if I want to rebuild just one specific module, I can't do `./helper.sh --build oms-statutory`, it'll rebuild the whole system. Flso there is no `docker-compose ps` like command.
We can add such commands later, but generally it'd be nice to have the ability to run something custom (though we should not abuse it and stick to the non-custom ones).

blacksph3re:
Exactly, other commands which are currently unsupported are scale, port, run and some other which I don't think we need atm. It's not for everyday's usage, but in a case of emergency when we need some custom command, it would be quite handy to use it

serge1peshcoff:
Actually one of the usecases I needed. I needed to drop postgresql container for statutory as I've messed it up and there was no way to do it other than copying that big command and adding what I needed to add (`rm -v postgres-oms-statutory`).

(on the local system ofc, not on prod :D)

serge1peshcoff (approved):

serge1peshcoff:
Rebased here: https://github.com/AEGEE/oms-docker/pull/95, so superseded.

PR #89, opened 2019-09-28
Chore/small stuff

linuxbandit:
Mostly update vagrant and eliminate dead code


base-docker-compose.yml
```diff
@@ -1,4 +1,4 @@
-version: "3.2"
+version: "2.4"
```
serge1peshcoff: will the docker-compose with multiple .yml files having different version work properly?
in general, I propose to have the same version everywhere

serge1peshcoff (approved):
LGTM, only thing unclear is that now we have different docker-compose versions everywhere

serge1peshcoff:
also nevermind CircleCI failing, I'll fix that in another PR (need to review it)

base-docker-compose.yml
linuxbandit: yes this is still to correct - it will be corrected in all the subrepos (or maybe i did already) but didn't update it here yet

serge1peshcoff: afaik it already has 3.4 in all of the other repos

serge1peshcoff: and also the answer to my question: no, it won't

linuxbandit: it does in the repo, but the PRs are addressing that (because as you said, it wouldn't work)

PR #98, opened 2019-11-11
Fixed traefik config generation

serge1peshcoff:
This should be the final step to allow us having the prod env by the way, didn't face any other issues

### DIFF
--- a/helper.sh
+++ b/helper.sh
@@ -35,7 +35,7 @@ init_boot ()
     touch "${DIR}"/secrets/acme.json # to avoid making it think it's a folder
     chmod 600 "${DIR}"/secrets/acme.json # Traefik doesn't let ACME challenge go through otherwise
 
-    envsubst < "${DIR}"/oms-global/traefik/traefik.toml.template > "${DIR}"/oms-global/traefik/traefik.toml
+    envsubst < "${DIR}"/oms-global/docker/traefik/traefik.toml.template > "${DIR}"/oms-global/docker/traefik/traefik.toml
 }
 
 # change passwords (currently deploy.sh [calls an external script])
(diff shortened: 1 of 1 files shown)


helper.sh
```diff
@@ -35,7 +35,7 @@ init_boot ()
     touch "${DIR}"/secrets/acme.json # to avoid making it think it's a folder
     chmod 600 "${DIR}"/secrets/acme.json # Traefik doesn't let ACME challenge go through otherwise
```
linuxbandit, line 37: Can you add the `touch` I mentioned? so we avoid having a folder

linuxbandit (changes requested):

helper.sh
serge1peshcoff: is there a point in `touch` ing the file and then overriding it with the correct one?

serge1peshcoff: it's not supposed to be not presented

helper.sh
```diff
@@ -35,6 +35,7 @@ init_boot ()
...
 
+    touch "${DIR}"/oms-global/docker/traefik/traefik.toml.template # to avoid making it think it's a folder
```
linuxbandit: I meant to touch the file not yet existing, i.e. `"${DIR}"/oms-global/docker/traefik/traefik.toml`

helper.sh
linuxbandit: it is, because we're talking about different files apparently :) oops

helper.sh
linuxbandit: that's where the misunderstanding came from :)

linuxbandit (changes requested):

helper.sh
serge1peshcoff: Whoops, my fault. I intended to touch the file and not the template, but just copied the wrong string lol.
Will fix today.

linuxbandit (approved):

PR #100, opened 2019-11-12
chore: add clarity on scripts, update readme

linuxbandit:
For nils

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,89 +1,125 @@
 # AEGEE-Europe's Online Membership System
-## OMS-docker
+## `oms-docker`
 
 ## Description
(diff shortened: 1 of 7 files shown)


serge1peshcoff (approved):

linuxbandit:
Do you think that what you suggested on slack about the clean start is understandable from this guide?

serge1peshcoff:
I'd say yes, except I don't see the `git submodule foreach git pull origin master` here for pulling the latest services dockerfile. other question is, given that the tag is always `latest` in docker-compose, do we ever need it?

linuxbandit:
the way it should be is that oms-docker versions are always updated with the latest module, so for every new release of a microservice, a `make bump` is called on this repo

linuxbandit:
I will merge it later today

PR #104, opened 2019-11-23
fix: greenlight the yamllint

linuxbandit:
i still miss a couple of things but the amount  of errors should be less


serge1peshcoff (approved):

linuxbandit:
will rebase

PR #105, opened 2019-11-24
fix: add newer version of cachet

linuxbandit:


oms-status/docker/cachet/.env.example
```diff
@@ -43,3 +43,8 @@ NEXMO_SMS_FROM=Cachet
 
 TRUSTED_PROXIES=
 
+SLACK_WEBHOOK_URL=
```
serge1peshcoff: do we have the provisioning for this file?

linuxbandit: we could, good idea

linuxbandit:
https://github.com/CachetHQ/Cachet/commit/6e9d3563ef7f76e40e28d09aa2a81a8c889586ee

comes from cachet issue 3441... apparently it's now broken because that link is a redirect to twitter and not a rss feed anymore..........

linuxbandit:
or not? apparently it is a cache (i.e. redis) problem, and dashboard can only be seen after `./helper.sh --docker -- exec cachet php artisan cache:clear`

serge1peshcoff (approved):

PR #108, opened 2020-02-29
feat: add observability of host

linuxbandit:


oms-monitor-agents/docker/docker-compose.yml
```diff
@@ -32,20 +32,42 @@ services:
     command: '--path.rootfs=/host'
     network_mode: "host"
     pid: "host"
+    # labels:
```
serge1peshcoff: why not exporting it through traefik?

linuxbandit:
Because it is on a different network (the host network), so traefik can’t route through it

Sent from mobile phone
________________________________

linuxbandit:
To* it

Sent from mobile phone
________________________________

serge1peshcoff:
@linuxbandit I'd rather let it be exposeable through traefik, but not on port 80, that'll give us benefits of using all of the traefik features (like putting basic auth in front of them)

linuxbandit:
I just said it's not possible :D they are on different subnetworks, so the (direct) routing won't work - however the metrics exposed by node-exporter are accessible through traefik.aegee.eu/ext-metrics/node-exporter, see  https://github.com/AEGEE/oms-docker/pull/108/files#diff-83d8c11b4a134b341c45587f720583d9R53

linuxbandit:
line 53 in case it's not shown properly (it is highlighted but the page opens slightly below)

serge1peshcoff (approved):

PR #109, opened 2020-03-02
feat: clone submodules for more thorough tests

linuxbandit:
Submodules were not loaded, thus the hadolint was always green for no reason


linuxbandit:
for best effectiveness of course the same circleci configuration should be also in every repo

serge1peshcoff (approved):

PR #116, opened 2020-03-12
Added .env vars for mailer

serge1peshcoff:
Fixes MEMB-826

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -57,6 +57,10 @@ PATH_OMS_SURVEY=oms-survey/docker/
 CORE_LOGIN=admin
 CORE_PASSWORD=5ecr3t
 
+SMTP_HOST=smtp.gmail.com
+SMTP_USER=admin@DOMAIN
+SMTP_PASSWORD=5ecr3t
+
 # Credentials external services
 USER_SENDGRID=
 PW_SENDGRID=5ecr3t


LeonVreling (approved):

.env.example
```diff
@@ -57,6 +57,10 @@ PATH_OMS_SURVEY=oms-survey/docker/
 CORE_LOGIN=admin
 CORE_PASSWORD=5ecr3t
 
+SMTP_HOST=smtp.gmail.com
+SMTP_USER=admin@aegee.eu
```
linuxbandit: can you change it to `admin@${BASE_URL}`?

linuxbandit (changes requested):

.env.example
serge1peshcoff: but it's the .env file, it won't work like that, we don't have envsubst here

linuxbandit: you're right, then change it to `admin@DOMAIN` such that email crawlers don't get this address

linuxbandit (approved):

.env.example
serge1peshcoff: @linuxbandit one argument for force-push: if I'd do it with another commit, then it'll stay in the history so people/robots/crawlers would be able to access it

PR #118, opened 2020-03-18
chore: tidy up prom config

linuxbandit:
new dashboard, and a lot more insight on how prom works

### DIFF
--- a/oms-monitor/docker/config/prometheus.yml
+++ b/oms-monitor/docker/config/prometheus.yml
@@ -18,63 +18,69 @@ rule_files:
 # Here it's Prometheus itself.
 scrape_configs:
   # The job name is added as a label `job=<job_name>` to any timeseries scraped from this config.
-  - job_name: 'local-diagnostics'
+  - job_name: 'scrape-app-diagnostics'
     # metrics_path defaults to '/metrics'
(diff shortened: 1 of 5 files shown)


oms-monitor/docker/config/prometheus.yml
```diff
@@ -18,63 +18,69 @@ rule_files:
...
     # scheme defaults to 'http'.
     static_configs:
       - targets:
-          - 'traefik:8080'
-          - 'cadvisor:8080'
-          - 'node-exporter-tunnel:5001'
           #  - 'statping:3000'
           - 'upmonitor:3000'
+          - 'prometheus:9090'
+          - 'grafana:3000'
```
serge1peshcoff, line 29: does Grafana expose metrics by itself?

serge1peshcoff (approved):

oms-monitor/docker/config/prometheus.yml
linuxbandit: yes, real question is if they are necessary - for now we should gather everything, then see if they are of use

serge1peshcoff: agree

PR #121, opened 2020-03-30
Feat/make launchers

linuxbandit:
still gotta test all the commands

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,25 +95,25 @@ remove-agents:
 
 # Backups
 backup_core:
-	bash ./helper.sh --execute postgres-oms-core-elixir -- pg_dump 'postgresql://postgres:$${PW_POSTGRES}@localhost/omscore_dev' --inserts > oms_core.sql.backup-$(date)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -16,14 +17,54 @@ start: #docker-compose up -d
...
+clean_docker_dangling_images:
+	docker rmi $(docker images -qf "dangling=true")
+
+clean_docker_images:
+	docker rmi $(docker images -q)
+
+clean: clean_docker_images clean_docker_dangling_images
+
+prune: clean
+	docker system prune -y
```
serge1peshcoff: `-a` maybe? this is what I do all the time

linuxbandit: that's not a good command imho, "too much" cleaning

linuxbandit: you do it only because we are forced to, being desperate for disk space

serge1peshcoff (approved):

linuxbandit:
tested commands

PR #127, opened 2020-04-15
chore: remove unused services

linuxbandit:
ciaohttp and cachet are replaced by statping


.env.example
```diff
@@ -2,13 +2,14 @@
...
 # Important variables
 BASE_URL=localhost.test
-
 HOSTNAME=
-EMAIL_DOMAIN=
 
 #we wish :(
 SHORTENER_DOMAIN=aeg.ee
 
+EMAIL_DOMAIN=
```
serge1peshcoff, line 10: what is this used for btw?

.env.example
```diff
@@ -20,59 +21,24 @@ BUGSNAG_KEY_EVENTS=123
...
-PATH_OMS_MONITOR=oms-monitor/docker/
-PATH_OMS_ELASTIC=oms-elastic/docker/
-PATH_OMS_WIKI=oms-wiki/docker/
-PATH_OMS_SURVEY=oms-survey/docker/
 
 CORE_LOGIN=admin
 CORE_PASSWORD=5ecr3t
 
-SMTP_HOST=smtp.gmail.com
+SMTP_HOST=smtp.gmail.com #!!!!
```
serge1peshcoff: why this?

.env.example
```diff
@@ -93,14 +59,47 @@ PW_STATPING=5ecr3t
 
 STATPING_TOKEN=
 
-PHP_APP_KEY=YdEgGZtx98WIBOKJky3gBvcnE+1ZW9+40HHSI0p+vWQ=
-RUBY_APP_KEY=YdEgGZtx98WIBOKJky3gBvcnE+1ZW9+40HHSI0p+vWQ=
+SSH_PORT=
+ELK_VERSION=
 
+#possibly remove?
```
serge1peshcoff: I think we can safely remove it (and wiki at all) and moving wiki plugin to other repo, as it's not the part of oms-docker and is hosted on zeus

.env.example
linuxbandit: I don't remember and i also needed to remove it because it could break the variable

.env.example
linuxbandit: yes but the objective imho is to bring it to us, we also can add automatic backups etc

linuxbandit: hm but ok i will remove it so it is in the same commit, we can look back here for the code

.env.example
linuxbandit: because BASE_URL will not always be our domain (especially now that i am making all the tests with barnabeu)

serge1peshcoff (approved):
lgtm

PR #128, opened 2020-04-15
feat: add mattermost chat

linuxbandit:


serge1peshcoff (approved):

.env.example
```diff
@@ -40,6 +40,8 @@ ELASTIC_PASSWORD=
 
 SLACK_WEBHOOK_GRAFANA=
 
+MM_USERNAME=
```
serge1peshcoff, line 43: can we rename it to MATTERMOST_USERNAME? it can seem unobvious what it does

linuxbandit:
we will not use mattermost

WikiRik:
@linuxbandit why not? Because we are using Slack more now?

linuxbandit:
correct, plus it still takes effort to administrate a self-hosted solution

serge1peshcoff:
but we'll still hit 10k messages limit right?

linuxbandit:
slack is for communication, not for archiviation

PR #132, opened 2020-04-23
Feat/greenlight shellcheck

linuxbandit:
for a mistake it is built on top of #131, so just merge that first


serge1peshcoff (approved):

helper.sh
```diff
@@ -220,11 +194,12 @@ if [[ "$#" -ge 1 ]]; then
             --docker) docker=true; ((command_num++)); shift ;;
             -v) verbose=true; shift ;;
 
-            --) shift ; arguments+=$@; break ;;
+            --) shift ; arguments+=("${@}"); break ;;
```
serge1peshcoff, line 197: one of the reasons to hate bash

serge1peshcoff: we've completely ditched it and replaced it with Python, it may be a good idea to do it as well

serge1peshcoff: (somewhere in the future, this is just a food for thought)

linuxbandit: agree, i was thinking about grunt (js). It is basically a replacement for make and for the helper script at once

serge1peshcoff: I'd rather go for gulp then, everybody uses it now

linuxbandit: > Each plugin for Gulp is written with the goal of doing one thing very well.

Unix philosophy? I'm in

linuxbandit: @WikiRik can you write it off as a low priority task?

serge1peshcoff: I've also used gulp before webpack (they are slightly different though: gulp is a task manager when webpack is a frontend build tool), worked flawlessly for me

WikiRik: > @WikiRik can you write it off as a low priority task?

What exactly?

PR #134, opened 2020-04-25
fix(logstash): fix config

serge1peshcoff:
1) removed `filter.prune`, it was cutting extra fields that we need
2) removed statutory from dissect, it's fully JSON now
3) spaces -> tabs
4) made config a little prettier

### DIFF
--- a/oms-monitor/docker/logstash/pipeline/logstash.conf
+++ b/oms-monitor/docker/logstash/pipeline/logstash.conf
@@ -1,55 +1,55 @@
 input {
-	tcp {
-		port => 5000
+  tcp {
+    port => 5000
     codec => json
-	}
-	udp {
-		port => 5000
+  }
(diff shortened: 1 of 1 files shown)


linuxbandit (approved):

oms-monitor/docker/logstash/pipeline/logstash.conf
```diff
@@ -79,28 +79,26 @@ filter {
 
 
 ## From logspout: our containers' output
```
serge1peshcoff, line 81: this block as a whole would be removed later in a future, once I'll switch all the services to log as json

serge1peshcoff:
@linuxbandit we also should add nginx parsing somewhere in the future (to parse oms-frontend, oms-events-static and oms-statutory-static)

PR #137, opened 2020-05-08
feat(deploy): added deploy script

serge1peshcoff:
1) set these .env files:

```
DEPLOY_HOST=<host of the server>
DEPLOY_PORT=<port>
DEPLOY_USER=<user>
DEPLOY_FOLDER=<folder>
DEPLOY_TOKEN=<Slack token>
```

2) run `./deploy.sh <service-name>`
3) enjoy

It pulls the container, then `up -d`'s it, then sends a Slack message (check out my thread).


WikiRik:
Can you add the variables to .env.example as well?

serge1peshcoff:
@WikiRik I thought about it but it's sensitive stuff so I decided not to.

serge1peshcoff:
@WikiRik then I did, but put some stub data instead, to not have our real data public.

WikiRik:
Yes, stub data is what I would have liked to see in there. Thanks for adding it!

deploy.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-export "$(grep -v '^#' ".env" | gxargs -d '\n')"
+export $(grep -v '^#' ".env" | gxargs -d '\n')
```
linuxbandit: is this gxargs a mac-specific command?

deploy.sh
```diff
@@ -0,0 +1,41 @@
...
+    exit 1
+fi
+
+SERVICE_NAME=$1
+
+echo "Pulling $SERVICE_NAME..."
+# shellcheck disable=SC2029
+ssh -p "$DEPLOY_PORT" "$DEPLOY_USER@$DEPLOY_HOST" "cd $DEPLOY_FOLDER && ./helper.sh --pull $SERVICE_NAME"
+
+echo "(Re)starting $SERVICE_NAME..."
```
linuxbandit, line 16: we could use directly this target (after changing it) https://github.com/AEGEE/oms-docker/blob/master/helper.sh#L244

we have to change it because it still has the former "we build on prod" mentality, tho that is the original purpose of the target

linuxbandit (changes requested):

deploy.sh
serge1peshcoff: yeah, but what  propose is more granular, it gives your more control on what you are  actually doing. Imagine somebody pulling  a container and not recreating it, then somebody trying to pull another container, then it'll recreate both of them and the deployer will be like "wtf just happened".

I'd say we should go for more granular commands and only deploy what's needed.

deploy.sh
serge1peshcoff: ah yes, I copied it from helper.sh (which I've changed but not committed) because xargs on mac doesn't have -d option and is not working correctly

deploy.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+export $(grep -v '^#' ".env" | gxargs -d '\n')
```
serge1peshcoff:
```suggestion
export $(grep -v '^#' ".env" | xargs -d '\n')
```

serge1peshcoff:
I actually have no idea if we need it, as we'll have aegee-bot in the future. I'll mark it as draft for now, let's see the future.

serge1peshcoff:
No need for this anymore, we have hubot now

PR #140, opened 2020-05-25
fix(logstash): improve provisioning

serge1peshcoff:
1) passing cloud.id and cloud.auth properly. Apparently commenting out these two line in logstash.yml helped
2) passing env vars to logstash from .env
3) using this vars in logstash.yml
4) no need to specify elastic host anymore in logstash pipeline
5) added docker-compose for dev setup so it'd work


serge1peshcoff:
disregard that, it won't work. I'll fix it today

PR #144, opened 2020-05-30
feat(logstash): write to different indexes based on tags

serge1peshcoff:
part 4 (should be the last one) of logstash pipeline improvements.

1) write images with tag `access-log`(traefik/nginx logs) to `access-log-${timestamp}`
2) write images with tag `bunyan-log` (app/logs) to `access-log-${timestamp}`
3) write other logs (mostly other services are these messages that couldn't be parsed with json parser) to `other-log-${timestamp}`

b1.barnab.eu uses this new schema right now, works okay.
only thing I don't know about is how to get rid of this copy-paste of host/username/password, but it may be not possible, so let's leave it like this.

### DIFF
--- a/oms-monitor/docker/logstash/pipeline/logstash.conf
+++ b/oms-monitor/docker/logstash/pipeline/logstash.conf
@@ -99,9 +99,28 @@ filter {
 }
 
 output {
-  elasticsearch {
-    hosts => "${ELASTIC_HOST}"
-    user => "${ELASTIC_USER}"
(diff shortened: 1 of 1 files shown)


serge1peshcoff:
also `logstash-*` index pattern is not used anymore, we can delete it I assume, also I've created 3 more index patterns (`access-log-*`, `bunyan-log-*`, `other-log-*`) and am changing my dashboards to use them.

linuxbandit (approved):

PR #146, opened 2020-06-08
fix(lint): greenify yamllint

serge1peshcoff:
1) not checking out submodules before running yamllint
2) fixed yamllint files, so it should be all green


linuxbandit (approved):
sweet

PR #148, opened 2020-06-22
feat(general): rewire submodules

serge1peshcoff:
1) re-wired all submodules with renamed repos
2) removed oms-core-elixir from setup, replaced it with core
3) changed default variables

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,6 @@
 [submodule "oms-global/docker/portal/files"]
   path = oms-global/docker/portal/files
   url = https://github.com/AEGEE/AEGEEPortal
-[submodule "oms-frontend"]
-  path = oms-frontend
-  url = https://github.com/aegee/oms-frontend
 [submodule "oms-mailer"]
 	path = oms-mailer
 	url = https://github.com/AEGEE/oms-mailer.git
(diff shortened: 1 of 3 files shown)


.gitmodules
```diff
@@ -1,21 +1,21 @@
-[submodule "oms-events"]
-  path = oms-events
-  url = https://github.com/AEGEE/oms-events.git
 [submodule "oms-global/docker/portal/files"]
   path = oms-global/docker/portal/files
```
WikiRik, line 2: Can you update these as well? Not sure if global is the proper word, do you have any other words that might replace this?

serge1peshcoff: I have no idea actually if we need it, @linuxbandit can you clarify?

serge1peshcoff: anyway I'd rather rename/remove it in different PR, as this is not the part of 'shortening the API prefixes' epic, I didn't touch it there

WikiRik:
Might be good to change the readme as well, most importantly the Installation section since it has links to the GitHub repo for cloning

.gitmodules
linuxbandit: indeed, postpone

linuxbandit (approved):

PR #151, opened 2020-07-03
fix: bootstrap procedure when not using start.sh

linuxbandit:
although it is very discouraged, it could happen of people
launching the command `vagrant up` before any initialisation


serge1peshcoff (approved):
I don't get what's going on here, but seems like it should be okay

PR #153, opened 2020-07-05
feat: API gateway for some architectural tests

linuxbandit:
It's interesting, play around it if you want. 

with this, the api will respond at api.${domain}/${servicename}. What is interesting is of course the plugins, because 'cute routes' we had them with traefik already


oms-global/docker/provision_kong.sh
```diff
@@ -0,0 +1,96 @@
...
+curl -i -X POST \
+  --url http://localhost:8001/services/discounts/routes \
+  --data 'paths[]=/discounts'
+
+
+### MAILER svc
+curl -i -X POST \
+  --url http://localhost:8001/services/ \
+  --data 'name=mailer' \
+  --data 'url=http://mailer:3000'
```
serge1peshcoff: It's 4000 iirc, also only /healthcheck is exposed

oms-global/docker/docker-compose.yml
```diff
@@ -40,9 +40,9 @@ services:
     expose:
       - "5432"
     environment:
-      POSTGRES_DB:       api-gw
-      POSTGRES_USER:     kong
-      POSTGRES_PASSWORD: ${PW_POSTGRES}
+      POSTGRES_DB:       api-gw # yamllint disable-line rule:colons
```
serge1peshcoff, line 43: if you want to ignore that rule, you can disable it globally

linuxbandit: nah it's that i like how this one is aligned

PR #154, opened 2020-07-24
feat(modules): add bump for single module

linuxbandit:
As per MEMB-344, we still do not know when to launch this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ build: #docker-compose build
 start: #docker-compose up -d
 	./helper.sh --start
 
-bootstrap: init build start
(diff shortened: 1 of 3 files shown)


linuxbandit:
according to this https://stackoverflow.com/a/9170206 (and to my tests back in the day) the function should realise there's something staged, and therefore checkout and commit. BUT IT FUCKING DOESN'T. It used to work, I remember. Can somebody review? 
it's the line 18 (and therefore 31 too)

linuxbandit:
nvm fuck this fucker

serge1peshcoff (approved):

PR #156, opened 2020-07-26
Feat/rename services

linuxbandit:
let's get rid of OMS where possible


.env.example
```diff
@@ -80,18 +80,16 @@ SUBDOMAIN_APIDOCS=apidocs.
...
 PATH_STATUTORY=statutory/docker
 PATH_FRONTEND=frontend/docker/
 PATH_DISCOUNTS=discounts/docker/
 PATH_MAILER=mailer/docker/
-PATH_OMS_SECURITY=oms-security/docker/
-PATH_OMS_SHORTENER=oms-shortener/docker/
-PATH_OMS_MONITOR=oms-monitor/docker/
+PATH_PASS_MANAGER=pass-manager/docker/
+PATH_MONITOR=monitor/docker/
 PATH_OMS_ELASTIC=oms-elastic/docker/
```
WikiRik, line 92: You forgot one

linuxbandit: good catch, I did not truly forget though, it will be part of another PR where we remove dead code (in this case, elastic is no more self-hosted but provided as donation, so I will remove the reference)

.env.example
```diff
@@ -10,7 +10,7 @@ EMAIL_DOMAIN=
 
 
 # To enable services, add them to the array, like
-ENABLED_SERVICES=oms-global:frontend:core:statutory:events:discounts:mailer
+ENABLED_SERVICES=gateways:frontend:core:statutory:events:discounts:mailer
```
serge1peshcoff, line 13: if I was a newbie, this would make me confused as from the name `gateway` it's not clear what does it do, maybe we can rename it to something more clear?

serge1peshcoff (commented):
check out my comment, otherwise lgtm

.env.example
linuxbandit: yeah i was also struggling to find a way to call it, the first thing that came to mind was "ingress", though it would still be unclear (or even less clear, as ingress is a 'keyword' from kubernetes). 

I mean, gateway has use in networking, ingress is a good English word yet possibly less known (and possibly mistaken for k8s stuff?)

What do you propose?

serge1peshcoff: is it only aegee portal or something more? if it is, we can name it `portal` or something

linuxbandit: it is traefik (and kong maybe), so it is a web (and an api) entry point to the application

WikiRik: @serge1peshcoff any suggestions on a different name? Otherwise I propose that we merge this and put in the readme what `gateways` contains

serge1peshcoff: yeah I've no idea on better naming, let's leave it as it is and rename later if needed.

serge1peshcoff (approved):

PR #159, opened 2020-09-17
fix(logstash): fix services logs processing

serge1peshcoff:
I've renamed the containers but did not update the logstash config, time to do so.


WikiRik:
Do we want to add stuff for the gsuite-wrapper as well?

serge1peshcoff:
I'm not sure about that, we may need to update gsuite-wrapper's logging for that. I'll check that later

linuxbandit:
why removing the stuff below? it was elixir logging way, which now is standardised?

serge1peshcoff:
@linuxbandit because core now logs in JSON, we don't need to parse it. only service writing in elixir format now is mailer

linuxbandit (approved):
thanks, thought so

PR #160, opened 2020-09-18
feat: add option to fast bootstrap, and to reset

linuxbandit:
@HassanCehef I can't find you as assignee

edit: nvm added the group everybody on the repo

The idea of this PR is 
1) give a 'fast lane' to people who would (like me) like to work on the infrastructure. They would need a working installation, but not in dev mode. Also by limiting the services enabled on startup (e.g. the `mailer` which is useless as it doesn't have a dev mode through sendgrid), building time would be reduced even in cases of dev mode (with a build)
2) give a chance to nuke vagrant, similarly to `helper.sh --nuke` but on the vagrant level instead of the docker level

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -8,9 +8,9 @@ SHORTENER_DOMAIN=aeg.ee
 
 EMAIL_DOMAIN=
 
-
-# To enable services, add them to the array, like
-ENABLED_SERVICES=gateways:frontend:core:statutory:events:discounts:mailer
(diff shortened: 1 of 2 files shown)


serge1peshcoff (approved):

HassanCehef (approved):
that part looks good. I'll check that the doc is in sync.

linuxbandit:
@HassanCehef the docs (PR #161) is in sync with the changes of this PR, FYI

PR #162, opened 2020-09-19
fix: the most important files are passed to d2u

linuxbandit:


vagrant-post-script/orchestrate_docker.sh
```diff
@@ -3,13 +3,12 @@
...
   make start
 else
   cp .env.example .env
-
-  #git reset --hard HEAD
-  #./helper.sh --bumpmodules
+  chmod +x helper.sh
+  dos2unix .env Makefile helper.sh
   make bootstrap
-  sleep 120 #to give time for the bootstrap
```
serge1peshcoff, line 14: we can also use `./helper.sh --wait-until-healthy` instead of it somewhere in the futuer

serge1peshcoff (approved):

vagrant-post-script/orchestrate_docker.sh
linuxbandit: i realised that the CLI is not responding anyway till it's finished (so, the message `finally` only appears afterwards anyway)

linuxbandit: basically it was only a way to not make people go on the browser till time had passed, it is not technical

PR #169, opened 2020-10-04
feat: differentiate between linux and mac

linuxbandit:
this is for a no-vagrant setup, it is bad bad bad to not use vagrant

@serge1peshcoff I assigned you because you should try it, I don't understand what's wrong with my (g)xargs. Probably not the gnu one, idk I have a lot of troubles with GNU tools even if I installed them explicitly (e.g. gsed etc)

### DIFF
--- a/helper.sh
+++ b/helper.sh
@@ -156,8 +156,13 @@ if [ ! -f "${DIR}"/.env ]; then #check if it exists, if not take the example
     cp "${DIR}"/.env.example "${DIR}"/.env
 fi
 # https://stackoverflow.com/questions/19331497/set-environment-variables-from-file-of-key-value-pairs
+if [[ $(uname) == "Linux" ]]; then
 # shellcheck disable=SC2046
-export $(grep -v '^#' "${DIR}/.env" | xargs -d '\n')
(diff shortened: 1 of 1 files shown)


serge1peshcoff (approved):

linuxbandit:
@serge1peshcoff have you tried it? I mentioned having problems with gxargs

serge1peshcoff:
I also have issues with quotes:

```
# in .env
TEST_VAR="contains qoutes"

# in .helper.sh
if [[ $(uname) == "Linux" ]]; then
# shellcheck disable=SC2046
  export $(grep -v '^#' "${DIR}/.env" | xargs -d '\n')
else #macos
# shellcheck disable=SC2046
  export $(grep -v '^#' "${DIR}/.env" | gxargs -0 -d '\n')
  echo "TEST_VAR $TEST_VAR"
fi

# output
$ ./helper.sh --list
./helper.sh: line 164: export: `qoutes"': not a valid identifier
TEST_VAR "contains
```

WikiRik:
@serge1peshcoff is that issue not just due to you having a space in the variable?

linuxbandit:
won't do, no vagrant = bad

PR #187, opened 2020-11-06
fix(ci): Remove git submodule from shell lint

WikiRik:
The shell linter did updates on the submodules, but we're updating the submodules regularly so it just slowing the check down imo


serge1peshcoff:
it also should be moved to submodules, e.g. each submodule should have its own shellcheck (probably I did it already)

serge1peshcoff (approved):

linuxbandit:
actually there's no shell at all in the subs, so it was a good removal

WikiRik:
In the docker folders there is bootstrap.sh right? Is that not taken into account?

linuxbandit:
only sometimes, and it is neglectable

PR #198, opened 2020-11-10
chore(dependabot): add dependabot config

WikiRik:


.github/dependabot.yml
```diff
@@ -0,0 +1,23 @@
+version: 2
+updates:
+  - package-ecosystem: "gitsubmodule"
+    directory: "/"
+    schedule:
+      interval: daily
+    open-pull-requests-limit: 99
+    reviewers:
+      - "WikiRik"
```
serge1peshcoff, line 9: you can also add me and Fabrizio here I think? either that or AEGEE/everybody group

WikiRik: I didn't want to do AEGEE/everybody so not to spam too much, but I can change it if you want

serge1peshcoff (approved):

PR #281, opened 2020-12-12
feat(backup): backup mgmt and notifications

linuxbandit:


scripts/notify.sh
```diff
@@ -0,0 +1,88 @@
...
+        -d "title=Backup completed" \
+        -d "body=The backup was a ${STATUS}" \
+      https://pushmeapi.jagcesar.se
+}
+
+function telegram_notifier(){
+  echo "not yet"
+}
+
+slack_notifier
```
serge1peshcoff, line 77: why these ones are called?

linuxbandit: just a quick-and-dirty way to enable or disable certain notifiers. In next iteration I'll actually make the function get a series of tokens so that more recipients can be configured for some means who allow this (e.g. slack is obviously only one, but when/if we add telegrams bot then more than one person can be notified).

This is of course a 'candy' thing, because if backups fail at 4 am we don't need to know right away (only slack is important), which is why I left it that way besides the improvement i mentioned above

WikiRik (approved):
Not sure if the notifications work, but since you tested it it should be fine

PR #418, opened 2021-03-20
fix: avoid building every time

linuxbandit:


linuxbandit:
hahaha same time of #417

PR #608, opened 2021-07-30
fix(.env): add LOGLEVEL

WikiRik:
See https://github.com/AEGEE/core/pull/327


WikiRik:
@serge1peshcoff @linuxbandit not sure if you have already seen this (and referenced) PR(s), but this should reduce the logging of the modules by quite a bit

serge1peshcoff (approved):

PR #759, opened 2021-11-18
chore(docker-compose): update docker-compose

WikiRik:
Already done in the microservices, but not here


WikiRik:
Okay, testing this locally meant also update Docker engine versions and that's not something we'll do now

PR #775, opened 2021-12-09
fix(bootstrap): only bootstrap the first time

WikiRik:
@linuxbandit this good? Or should we do something else in the Makefile depending on the MYAEGEE_ENV in the .env file?

### DIFF
--- a/vagrant-post-script/orchestrate_docker.sh
+++ b/vagrant-post-script/orchestrate_docker.sh
@@ -11,7 +11,7 @@ chmod +x helper.sh
 find . -type f -name "*.sh" -exec dos2unix {} \+;
 dos2unix .env Makefile
 
-if [[ -f .env ]]; then
+if [[ -f .init ]]; then
   make start
 else
   make bootstrap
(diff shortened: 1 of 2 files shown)


WikiRik:
Okay, I've updated the MYAEGEE_ENV because I think for the longer run it is better to build all containers locally immediately in case of any errors. Will take more time on the first thing, but not on the longer run

vagrant-post-script/orchestrate_docker.sh
```diff
@@ -11,7 +11,7 @@ chmod +x helper.sh
 find . -type f -name "*.sh" -exec dos2unix {} \+;
 dos2unix .env Makefile
 
-if [[ -f .env ]]; then
+if [[ -f .init ]]; then
```
linuxbandit, line 14: the file .init is never created anywhere

linuxbandit:
> Okay, I've updated the MYAEGEE_ENV because I think for the longer run it is better to build all containers locally immediately in case of any errors. Will take more time on the first thing, but not on the longer run

hmmm i am on the fence for this

vagrant-post-script/orchestrate_docker.sh
WikiRik: https://github.com/AEGEE/MyAEGEE/blob/master/helper.sh#L74 as part of make init

WikiRik:
I understand, and maybe we can find out an intermediate solution soon. But currently the initial setup takes too long with the core seeding

WikiRik:
@linuxbandit I would prefer to merge this before the hackathon, are you okay with me merging this? We'll find out a better solution later

linuxbandit (approved):

PR #854, opened 2022-10-31
chore(toolings): add python notifier helper

linuxbandit:
A rewrite of the shell function in python as PoC of having our toolings in Python. 

Some notable mentions:
- compared to the sh, `pushme` disappeared so we added logsnag instead (a cool SaaS)


WikiRik:
Any reason why you did not remove the bash file in this PR?

linuxbandit:
> Any reason why you did not remove the bash file in this PR?

mainly i forgot, fixed and i even "plugged it in"

scripts/notify.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+#the 4 notification channels:
+# - slack
+# - notifi (works for both phone and mac)
+# - push on the phone (via 'logsnag', pushme disappeared)
+# - telegram (wip)
+
+
+token_file = f'{os.path.realpath(os.path.dirname(__file__))}/../secrets/tokens.json'
```
WikiRik, line 30: In the bash file we have an example of this file, do we want that here as well? 

```
# example of tokens.json:
# {
#   "slack": "123",
#   "notifi": "456",
#   "pushme": "789"
# }
```

Personally I'm a bit on the fence since it is quite a logical setup for the file

scripts/notify.py
```diff
@@ -0,0 +1,93 @@
...
+import requests
+import datetime
+from logsnag import LogSnag
+import slack_notifications as slack
+
+
+#the 4 notification channels:
+# - slack
+# - notifi (works for both phone and mac)
+# - push on the phone (via 'logsnag', pushme disappeared)
```
WikiRik:
```suggestion
# - push on the phone (via 'logsnag')
```

Might be good to mention it in the PR description, but I don't think it's necessary here

scripts/notify.py
```diff
@@ -0,0 +1,93 @@
...
+import sys
+import requests
+import datetime
+from logsnag import LogSnag
+import slack_notifications as slack
+
+
+#the 4 notification channels:
+# - slack
+# - notifi (works for both phone and mac)
```
WikiRik:
```suggestion
# - notifi (works for iOS/Android but also MacOS and Linux)
```

Just to specify the supported platforms more

scripts/notify.py
```diff
@@ -0,0 +1,93 @@
...
+      footer='Hope the backup is never needed',
+      color='#37FDFC' if STATUS == "SUCCESS" else "#ff5a36",
+      fields=[
+        slack.Attachment.Field(
+            title='Completed at:',
+            value=f'{datetime.datetime.now().strftime("%d %b %Y, %H:%M:%S %Z")}',
+            short=False
+        ),
+      ],
+      #fallback=EVENT # missing
```
WikiRik: Missing from where?

scripts/notify.py
```diff
@@ -0,0 +1,93 @@
...
+  api_key = tokens["slack"]
+
+  slack.ACCESS_TOKEN = api_key
+
+  attachment = slack.Attachment(
+      title='Backup result',
+      author_name=f'Host: {os.uname()[1]}',
+      text=EVENT,
+      footer='Hope the backup is never needed',
+      color='#37FDFC' if STATUS == "SUCCESS" else "#ff5a36",
```
WikiRik:
```suggestion
      color='#37FDFC' if STATUS == "SUCCESS" else "#FF5A36",
```
Just for consistency

scripts/notify.py
```diff
@@ -0,0 +1,93 @@
...
+
+    icon = "🎉" if STATUS == "SUCCESS" else "❌"
+    notify = False if STATUS == "SUCCESS" else True
+
+    logsnag.publish(
+        channel="spare", #FIXME: only works for fab
+        event=EVENT,
+        icon=icon,
+        notify=notify
+    )
```
WikiRik, line 76:
```suggestion
    logsnag.publish(
        channel="spare", #FIXME: only works for fab
        event=EVENT,
        icon="❌",
        notify=True
    )
```

We're only alerting in case of a failure, or do we want to keep this for testing purposes?

scripts/notify.py
```diff
@@ -0,0 +1,93 @@
...
+  for token in tokens["notifi"]:
+
+    response = requests.post('https://notifi.it/api', {
+        'credentials': token,
+        'title': EVENT,
+        # one can also add 'message' 'link' and 'image'
+    })
+
+    if response.status_code != 200:
+        raise Exception(response.status_code, response.text)
```
WikiRik, line 89: Is this built in for Slack and Logsnag? Or do we need to do error handling for those as well?

WikiRik (commented):
General comment for later; we should probably also add a Python linter to CircleCI for this.

Overall looks good, just some nit picky stuff.

scripts/notify.py
linuxbandit: the library doesn't provide that same parameter - idk what it meant though and I didn't want to check on slack API

scripts/notify.py
linuxbandit: I think it's better to alert "extra" in case of failures (also because there are a certain number of events for free in logsnag, and Idk about notifi that well -- one thing for sure, i'd like to start up our own instance of self-hosted, maybe ntfy.sh)

scripts/notify.py
linuxbandit: all in the library

scripts/notify.py
WikiRik: Let's remove it then

WikiRik: Nvm, already done. Thanks!

scripts/notify.py
WikiRik: Sure. Let's leave this here for now then

WikiRik (approved):

PR #857, opened 2022-10-31
feat: add bootstrap files for python

linuxbandit:


vagrant-post-script/setup_python.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+# call it with
+#   ./setup_python.sh venv
+# (or just any argument) will have it installed for local dev
```
WikiRik, line 5: What's the difference?

linuxbandit: I always did it with virtual env because it's the good way, but i realised that idk how to call the script from dump.sh without calling the activation of the virtual env so I went the secure way and installed the dependencies globally

linuxbandit: in other words, a way to see it is that one is for local dev and one is for the server

WikiRik: So with virtual env you basically install the dependencies in the folder itself and otherwise you install it for the entire server. Did I understand that correctly?

linuxbandit: yes, it's similar to nvm for node, where to avoid dependency hell or bad compatibility you install a specific version not globally

WikiRik (approved):

PR #858, opened 2022-10-31
feat: add self-hosted notification system

linuxbandit:


linuxbandit:
I am not closing it yet, but it became much less priority than the dispatcher

linuxbandit:
this would work better with ssl -> traefik2 -> monorepo

PR #1086, opened 2023-07-08
feat: add some improvements for bootstrap/reset

linuxbandit:
- install frontend stuff automatically
- auto-delete .init file on vagrant destroy


vagrant-post-script/orchestrate_docker.sh
```diff
@@ -16,3 +16,10 @@ if [[ -f .init ]]; then
 else
   make bootstrap
 fi
+
+
+# frontend helper
+cd frontend || exit 1
+echo "node: $(node --version)"
+echo "npm: $(npm --version)"
+npm i || npm i --force && npx vue-cli-service build && chmod 775 -R dist && docker restart myaegee_frontend_1
```
WikiRik, line 27:
```suggestion
npm i || npm i --force && npm run build && chmod 775 -R dist && docker restart myaegee_frontend_1
```

What does `npm i || npm i --force` do? Does it try `npm i` first and if that fails use `npm i --force`?

Vagrantfile
```diff
@@ -11,17 +12,20 @@ Vagrant.configure("2") do |config|
...
+  config.vm.network :private_network, ip: "192.168.168.168"
+  ## Port forwarding
+  #NOTE: there could be a different script that sets the resolv.conf and then
   # calls vagrant up
-  #SSH from anywhere on the network (sshd)
-  config.vm.network :forwarded_port, guest: 22, host: 2222, host_ip: "127.0.0.1", id: "ssh", auto_correct: true
-  #In case somebody does not use "appserver" but "localhost"
-  config.vm.network :forwarded_port, guest: 80, host: 8888, id: "main"
-
+  #If you want to SSH from anywhere on the network (sshd) uncomment this
```
WikiRik, line 23: Why not do this by default? Can it conflict with other applications?

Vagrantfile
```diff
@@ -11,17 +12,20 @@ Vagrant.configure("2") do |config|
     ]
   end
 
-  config.vm.define "appserver"
+  config.vbguest.auto_update = false
```
WikiRik, line 15: What does this do?

vagrant-post-script/orchestrate_docker.sh
linuxbandit: correct. i kept using npx because all of cristian's problems are not encountered here (damn), what do you think should be used?

Vagrantfile
linuxbandit: no, it's just that it is uncommon to want to be on a machine A and ssh to vagrant hosted on machine B

Vagrantfile
linuxbandit: every boot, at least for my machine, tries to upgrade (and fail) the guest additions from 7.0.6 to 7.0.8, so I am trying to install it at the first bootstrap (see start.sh) and never again

vagrant-post-script/orchestrate_docker.sh
WikiRik: The `npx` one doesn't work with Node 18 so we should use `npm run build`. Before it was both executing the same command

linuxbandit: BUT we do use node 16 inside vagrant (see prerequisites.sh)

WikiRik: We can update that btw. All submodules use Node 18 now

vagrant-post-script/prerequisites.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-sudo apt-get update
-sudo apt-get install curl -y
+sudo apt-get update > /dev/null
+sudo apt-get install curl -y > /dev/null
 
 NODE_VERSION=16
```
WikiRik, line 6:
```suggestion
NODE_VERSION=18
```

vagrant-post-script/orchestrate_docker.sh
linuxbandit: iirc there was a reason for using 16, given that this pinning happened only 2 months ago. Of course I do not remember this reason :))) if you say to put 18, we put 18

WikiRik: That was because frontend did not work with Node 18 yet, but that was fixed last week. We should be good for Node 18

linuxbandit: the pain never ends https://stackoverflow.com/questions/74815894/install-nodejs-v18-on-ubuntu-18-04

WikiRik: Maybe we can have Ubuntu 20.04 (or 22.04) locally already? I used 20.04 for my own server and that works great, now I am on 22.04 but I didn't do much on it yet

linuxbandit: I would keep it for a separate MR

vagrant-post-script/prerequisites.sh
linuxbandit: see my last comment

WikiRik (approved):

PR #1087, opened 2023-07-09
feat: bump versions of ubuntu, docker engine, node

linuxbandit:
Marked as draft as this can't be tested for now - see slack (if it disappears: there are bootstrap problems even in the current version, and the upgrade isn't enough to solve them)

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -47,7 +47,7 @@ else
   if ( $fast ); then
     sed -i 's/development/production/' .env
   fi
-  vagrant box list | grep "202303.13.0" -q || vagrant box add bento/ubuntu-18.04 --provider virtualbox --box-version "202303.13.0" -c
+  vagrant box list | grep "202303.13.0" -q || vagrant box add bento/ubuntu-22.04 --provider virtualbox --box-version "202303.13.0" -c
   vagrant plugin list | grep vbguest -q || vagrant plugin install vagrant-vbguest
   vagrant up
 fi
--- a/vagrant-post-script/install_docker.sh
+++ b/vagrant-post-script/install_docker.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-DOCKERVERSION="19.03.1"
+DOCKERVERSION="24.0.2"
 
 sudo dpkg -l | grep docker && { echo "[Vagrant] ###################     Docker already installed, exiting"; exit; }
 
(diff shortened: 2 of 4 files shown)


vagrant-post-script/install_docker.sh
```diff
@@ -15,7 +15,7 @@ sudo apt-key fingerprint 0EBFCD88 || { echo "[Vagrant] ###################     F
 sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu bionic stable"
```
WikiRik, line 15: This should be updated as well, just like all other instances of `bionic`

WikiRik (commented):
I don't recognize the issues you're running into for Node but I'll look into it later. I did see a SSH message, did you try it out on the code before the merge of your previous PR?

vagrant-post-script/install_docker.sh
linuxbandit: good catch

vagrant-post-script/prerequisites.sh
```diff
@@ -3,8 +3,8 @@
 sudo apt-get update > /dev/null #avoid spamming the output
 sudo apt-get install curl htop -y
 
-NODE_VERSION=16
-NPM_VERSION=9.6.4
+NODE_VERSION=18
+NPM_VERSION=9.8.8
```
WikiRik, line 7:
```suggestion
NPM_VERSION=9.8.0
```
This is the latest version of NPM

start.sh
```diff
@@ -47,7 +47,7 @@ else
   if ( $fast ); then
     sed -i 's/development/production/' .env
   fi
-  vagrant box list | grep "202303.13.0" -q || vagrant box add bento/ubuntu-18.04 --provider virtualbox --box-version "202303.13.0" -c
+  vagrant box list | grep "202303.13.0" -q || vagrant box add bento/ubuntu-22.04 --provider virtualbox --box-version "202303.13.0" -c
   vagrant plugin list | grep vbguest -q || vagrant plugin install vagrant-vbguest
```
WikiRik, line 51: Can you add `vagrant plugin install vagrant-vbguest` to the installation instructions for non-linux? Since they do not use start.sh but go to use `vagrant up` immediately. I just noticed it when booting things up from scratch on my Windows laptop

WikiRik:
Don't forget to update orchestrate_docker.sh to use the Node 18 frontend command instead of the Node 16 one.

WikiRik:
EDIT2: this was the first time in a few years that I've bootstrapped this from scratch so I didn't see some issues that we've had. statutory and mailer were the two that didn't start properly but I fixed them

vagrant-post-script/install_docker.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-DOCKERVERSION="19.03.1"
+DOCKERVERSION="24.0.2"
```
WikiRik, line 3: New thing which can also be solved in a future PR since it's recommended to update the docker-compose yaml in the various submodules as well; we should upgrade to docker compose v2. See https://docs.docker.com/compose/migrate/

linuxbandit:
Turns out it's better to use 20.04 and not 22.04 for some node and docker things.. but that's only preliminary observation and can't reproduce yet

WikiRik:
I'm fine with updating to 20.04 first and 22.04 later

linuxbandit:
- I implemented your advice of `vagrant plugin install vagrant-vbguest` into #1223 
- I want to improve the install of docker and the rest via reusing the ansible manifest that we use in prod, and we also got to understand that it's better to bump to ubuntu 20.04 (at least from my experimental branch, but as i mentioned, i still gotta reproduce).

because of that, i am closing the PR with the intent of working on it after our host migration, and via ansible

WikiRik:
All good!

PR #1205, opened 2023-09-16
chore: small infra improvements

linuxbandit:
work on backups
work on script location
add pyenv


WikiRik:
You should update https://github.com/AEGEE/MyAEGEE/blob/stable/.circleci/config.yml#L20 as well so pylint can check that folder

scripts-server/restore.sh
```diff
@@ -4,6 +4,27 @@
 # Before executing make sure nobody is accessing the databases, i.e. stop all services except for the actual databases
 
 postgres_hosts=(postgres-core postgres-oms-statutory postgres-oms-events)
+#postgres_hosts=($2) #FIXME still to test
+
+#to get postgres version programmatically (wip): docker ps | grep postgres | awk '{print $2}'
+# NB weak because every ms may have a different version of postgres
```
WikiRik, line 10: We should use the same version for all microservices though so would be worth to make some PRs if this is not the case right now

scripts-vagrant_provision/prerequisites.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+sudo apt-get update
+# For NVM
+sudo apt-get install -y curl
+# For pyenv
+sudo apt-get install -y git build-essential libssl-dev zlib1g-dev libbz2-dev libreadline-dev libsqlite3-dev wget curl llvm libncurses5-dev libncursesw5-dev xz-utils tk-dev libffi-dev liblzma-dev python-openssl
+
+NODE_VERSION=16
```
WikiRik, line 9:
```suggestion
NODE_VERSION=18
```
16 is EOL already since this week

WikiRik (commented):
Nice stuff! Haven't tested anything yet but I don't see anything weird. I do have two suggestions already

scripts-vagrant_provision/prerequisites.sh
linuxbandit: yes but until we update to 20.04 I think it is better to keep it this way

scripts-server/restore.sh
linuxbandit: you are suggesting to also update all the postgres in the ms? I would not do that, it takes longer (more PRs). in that regard, I want to explore the monorepo way, because i also have the upcoming upgrade to traefik2 that would use a shortcut..!

WikiRik: If you can't find it on Slack, this is an article I found that might be a nice way to convert to monorepo; https://x3ro.de/integrating-a-submodule-into-the-parent-repository/

linuxbandit: This isn't going to happen in this pr anyway, can you solve the conversation and approve?

WikiRik (approved):

PR #1221, opened 2023-09-28
fix: orchestrate_docker was not changing fe perm

linuxbandit:
further add variables on vagrantfile

### DIFF
--- a/README.md
+++ b/README.md
@@ -366,7 +366,7 @@ This file which is located on the _HOST_ machine is used to add new [subdomains 
 
 - Remember to check what is running on your _GUEST_ via `docker ps` (or portainer). Enter your _GUEST_ with `vagrant ssh` (from your _HOST_).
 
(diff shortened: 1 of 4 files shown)


WikiRik (approved):

WikiRik:
We'll need a similar fix for non-Vagrant users, right?

linuxbandit:
similar in which way? pin of docker?

Generally I don't want to support other unofficial methods when one is advised. As long term solution what should happen truly is an ansible playbook that provisions whatever one wants: be it the local machine, the local VM, or the remote server

WikiRik:
The frontend permissions, or this only for the very initial build of frontend?

linuxbandit:
the frontend fix is for the first run in development mode, via vagrant; in the readme I could add in the 'troubleshoot' section a note about the 403, then i will push and merge

WikiRik (approved):

PR #1223, opened 2023-09-29
feat: add small diagnostics output

linuxbandit:
Also add note about GNU grep that should close #169

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,7 @@ For any troubleshoot, see [Advanced URL mapping and troubleshoot](#advanced-url-
 
 Once set up the mapping, you can continue the installation:
 ```
+vagrant plugin install vagrant-vbguest
 vagrant up
 ```
 
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Install
 
 Even if you have a linux box, this is **very** recommended. If you decide to not do it, *sigh...* but don't come to cry to us.
 
-If you decide you know better than us, [install docker and docker-compose](https://docs.docker.com/compose/install/) on your Windows/Linux/Mac machine, instead of Virtualbox and Vagrant. (Make sure you install the correct versions: tested with Docker CE 19.03.1 and docker-compose 1.24.1 --  also, [Mac and linux have different versions of grep](https://stackoverflow.com/a/59393993) so again, your problem ;-) )
+If you decide you know better than us, [install docker and docker-compose](https://docs.docker.com/compose/install/) on your Windows/Linux/Mac machine, instead of Virtualbox and Vagrant. (Make sure you install the correct versions, they can be found in the provisioning scripts --  also, [Mac and linux have different versions of grep](https://stackoverflow.com/a/59393993) so again, your problem ;-) )
 
 Note: if you use Vagrant, Docker will be already automatically on the virtual machine.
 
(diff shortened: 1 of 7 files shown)


helper.sh
```diff
@@ -112,6 +112,39 @@ compose_wrapper ()
     return $?
 }
 
+summary_images ()
+{
+    service_string=$(printenv ENABLED_SERVICES)
+    own_services=( "frontend" "core" "events" "statutory" )
```
WikiRik:
```suggestion
    own_services=( "frontend" "core" "events" "statutory" "summeruniversity" "network" "discounts" "mailer" "gsuite-wrapper" )
```

WikiRik (approved):

helper.sh
```diff
@@ -234,6 +269,11 @@ if (( ${command_num} > 1 )); then
     echo "Usage: helper.sh {--init|--build|--start|--refresh|--monitor|--stop|--down|--restart|--nuke|--execute|--bump|--docker} [-v]"; exit 1
 fi
 
+if [[ $HOSTNAME != "appserver" ]]; then
```
WikiRik: Does this work in a non-vagrant environment?

linuxbandit: no, but as per the other comments yesterday I have no more energy to support non-vagrant - and after all, it's very clear in the readme that non-vagrant shouldn't be the case

linuxbandit: this line has to be edited further, though

WikiRik: "It works fine on my machine", but I understand

linuxbandit: what do you mean? vagrant is exactly to avoid the "works on my machine"

WikiRik: If it works for people with Docker and we don't have to give active support, it's fine what people use imo. Or at least; I'm using the no-vagrant version for my server. No need to update this though, I can always comment this out on my machine. 
This guard will help more people than it will cost me time, so it's fine this way

README.md
```diff
@@ -27,7 +27,7 @@ Install
 
 Even if you have a linux box, this is **very** recommended. If you decide to not do it, *sigh...* but don't come to cry to us.
 
-If you decide you know better than us, [install docker and docker-compose](https://docs.docker.com/compose/install/) on your Windows/Linux/Mac machine, instead of Virtualbox and Vagrant. (Make sure you install the correct versions: tested with Docker CE 19.03.1 and docker-compose 1.24.1)
+If you decide you know better than us, [install docker and docker-compose](https://docs.docker.com/compose/install/) on your Windows/Linux/Mac machine, instead of Virtualbox and Vagrant. (Make sure you install the correct versions: tested with Docker CE 19.03.1 and docker-compose 1.24.1 --  also, [Mac and linux have different versions of grep](https://stackoverflow.com/a/59393993) so again, your problem ;-) )
```
WikiRik:
```suggestion
If you decide you know better than us, [install docker and docker-compose](https://docs.docker.com/compose/install/) on your Windows/Linux/Mac machine, instead of Virtualbox and Vagrant. (Make sure you install the correct versions, they can be found in the provisioning scripts --  also, [Mac and linux have different versions of grep](https://stackoverflow.com/a/59393993) so again, your problem ;-) )
```
People can figure the tested versions out on their own, saves us maintenance in the future

WikiRik (approved):

README.md
linuxbandit: can you add "the versions are easy to find in the provisioning scripts" ?

WikiRik: Done

PR #1238, opened 2023-10-27
fix: correct regex that prevents startup

linuxbandit:
also add proper "~" expansion
also add task in the 'suggested first tasks'


WikiRik (approved):

WikiRik:
@linuxbandit I'll let you merge this one

PR #1254, opened 2024-01-13
fix: fix typo in start.sh usage

WikiRik:
@linuxbandit not sure how to deal with the TODO since I think we'll need to pass the no-vagrant from start.sh to helper.sh. Or is there some kind of global thing we can set? I'm not really knowledgeable about shell things

### DIFF
--- a/helper.sh
+++ b/helper.sh
@@ -189,6 +189,10 @@ function retry {
 }
 
 # TODO: ignore this when using --no-vagrant in start.sh
+if [[ ! $HOSTNAME =~ ^appserver ]]; then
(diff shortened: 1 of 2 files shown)


linuxbandit (approved):

linuxbandit:
tbh i think we should remove the option of no-vagrant. it was a "concern" (imho tantrum) from nico, for now let's go with this and thanks for the PR! lmk your comment

WikiRik:
I think that later with GitHub Codespaces (and right now on my own server) it's easier to not use vagrant since we can just destroy the Codespace and generate a new one if things are really broken. It feels weird to me to use another layer of virtualization in that case.
But we can take a look at this after https support has been merged etc. I can always manually run some commands to circumvent usage of vagrant

PR #1271, opened 2024-01-21
feat: use ansible provisioner

linuxbandit:
for now there will be a duplication of the logic between infrastructure and this ansible, but it will be harmonised in successive PRs for both repos


scripts-vagrant_provision/local.yml
```diff
@@ -0,0 +1,12 @@
+---
+nodejs_version: "16.20.2"
+npm_version: "9.6.4"
+
+python_version: "3.10.3"
+
+docker_version: "19.03.1"
```
WikiRik: Can we use 19.03.5? See https://github.com/timescale/timescaledb/issues/5414

It's an error that Cristian got as well

scripts-vagrant_provision/diagnostics.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+
+ALMOST_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+DIR="${ALMOST_DIR}/.."
+export $(grep -v '^#' "${DIR}/.env" | xargs -d '\n')
+
+summary_images ()
+{
+    service_string=$(printenv ENABLED_SERVICES)
+    own_services=( "frontend" "core" "events" "statutory" )
```
WikiRik, line 11: This should include the others as well, right? Discounts, network, gsuite-wrapper, summeruniversity, mailer. Possibly knowledge but we'll likely remove that soon

scripts-vagrant_provision/provision.yml
```diff
@@ -0,0 +1,175 @@
...
+        path: "{{ item }}"
+        state: directory
+        owner: vagrant
+        group: vagrant
+        recurse: yes
+      loop:
+        - /opt/myaegee
+        - /opt/backups
+
+    - name: Check current Nodejs version
```
WikiRik, line 42: Why would we check the version here if we're checking in the next step that it's not installed?

scripts-vagrant_provision/provision.yml
```diff
@@ -0,0 +1,175 @@
...
+###
+- name: Create banners for connection
+  hosts: all
+  become: true
+  tags: ['motd']
+
+  tasks:
+      - name: Ensure a locale exists
+        locale_gen:
+          name: en_IE.UTF-8
```
WikiRik: Not that it really matters, but why Ireland?

WikiRik (commented):
Interesting stuff! I'll do some tests on my own laptop later but I do have a few questions/remarks from looking at the code

scripts-vagrant_provision/diagnostics.sh
linuxbandit: yes but that part is still a prototype so I won't add them all for now

scripts-vagrant_provision/provision.yml
linuxbandit: if it's our version, we don't uninstall it
if it's another version, we completely uninstall (and let the role take care of it)
in fairness, if the role is done properly, they should already check if there's a node and uninstall it, so maybe i will remove this

scripts-vagrant_provision/provision.yml
linuxbandit: it's the locale I have, US is shit and GB has something that I don't remember (the reason I have is a bonus for italian speakers: gaelic has accents éáó so I can use them)

scripts-vagrant_provision/provision.yml
linuxbandit: after checking the role, it's better if it stays

WikiRik:
We'll need to update the README as well. I got this error;
```
Windows is not officially supported for the Ansible Control Machine.
Please check https://docs.ansible.com/intro_installation.html#control-machine-requirements
    appserver.test: Running ansible-playbook...
The Ansible software could not be found! Please verify
that Ansible is correctly installed on your host system.

If you haven't installed Ansible yet, please install Ansible
on your host system. Vagrant can't do this for you in a safe and
automated way.
Please check https://docs.ansible.com for more information.
```

Will update once I have it installed.

EDIT: Or not. The link mentioned above is not available anymore, but I think this is similar; https://docs.ansible.com/ansible/latest/installation_guide/installation_distros.html#installing-ansible-on-windows
So Ansible is not supported on Windows. WSL could be used, but that's a bigger impact. We might need to re-evaluate if we want this

WikiRik (approved):
Don't have a good setup right now where I can test on Linux/Mac but in theory this should work

PR #1287, opened 2024-03-09
feat: add traefik2 for all the services (except monitoring) in dev mode

linuxbandit:
Currently the traefik2 configuration is for DEV, i.e. it assumes a bunch of self-signed certificates (via mkcert) to be provisioned. In other words there is not yet the ACME http01 challenge. It will be introduced after i make sure it works

NOTE: this MR is NOT against `stable`


gsuite-wrapper/docker/docker-compose.yml
```diff
@@ -22,11 +25,19 @@ services:
...
       # BUGSNAG_KEY: "${BUGSNAG_KEY_GSUITE_WRAPPER}"
     labels:
-      - "traefik.backend=gsuite-wrapper"
-      - "traefik.port=8084"
-      - "traefik.hc.frontend.rule=Path:/api/gsuite-wrapper/healthcheck;PathPrefixStrip:/api/gsuite-wrapper"
-      - "traefik.doc.frontend.rule=Path:/api/gsuite-wrapper/api-docs.json;PathPrefixStrip:/api/gsuite-wrapper"
-      - "traefik.enable=true"
+      #- "traefik.hc.frontend.rule=Path:/api/gsuite-wrapper/healthcheck;PathPrefixStrip:/api/gsuite-wrapper"
+      #- "traefik.doc.frontend.rule=Path:/api/gsuite-wrapper/api-docs.json;PathPrefixStrip:/api/gsuite-wrapper"
+      # ---------old ^ for some reason it was all that was exposed, TODO: check
```
WikiRik, line 30: Are you sure that's all what was exposed? I thought we used gsuite-wrapper on prod once, since it is deployed there

knowledge/docker/docker-compose.yml
```diff
@@ -10,6 +10,9 @@ services:
         environment:
             POSTGRES_USER: "postgres"
             POSTGRES_PASSWORD: "${PW_POSTGRES}"
+        labels:
+            # we don't access via traefik
+            - "traefik.enable=false"
 
     knowledge:
```
WikiRik, line 17: Not sure if we should have this file here. We never actually got a use for it and are using Confluence now for KMS. Might be good to remove it when moving to monorepo and just archive the old repo.

WikiRik (commented):
Looks good!

events/docker/docker-compose.yml
```diff
@@ -37,25 +40,22 @@ services:
...
+            - "traefik.http.routers.events.middlewares=events-stripprefix"
+            - "traefik.http.middlewares.events-stripprefix.stripprefix.prefixes=/api/events"
+
+            - "traefik.http.routers.metrics.rule=Host(`${SUBDOMAIN_FRONTEND}${BASE_URL}`) && PathPrefix(`/api/events/metrics`)"
+            - "traefik.http.routers.metrics.middlewares=events-stripprefix, metricsauth"
 
     events-static:
         restart: on-failure
-        image: aegee/nginx-static:latest
+        image: aegee/nginx-static:prod
```
WikiRik, line 56: Does this mean we also updated nginx versions? Not sure if we really maintained this image. Which is the Dockerfile for it?

knowledge/docker/docker-compose.yml
linuxbandit: By `this file` you mean this module altogether I presume?

WikiRik: Yep

gsuite-wrapper/docker/docker-compose.yml
linuxbandit: well weirdly, yes, but again that's why i wrote i have to double check

events/docker/docker-compose.yml
linuxbandit: i only made a Dockerfile FROM aegee/nginx-static:latest and with the 3 COPY instructions, and pushed

PR #1314, opened 2024-04-27
chore: add a variable for the name of the email sender

CASician:
When receiving an email from mailer@aegee.eu, the resulting name of the sender will be stored in this variable.


CASician:
A second pull request will be linked to this, but in mailer repo.

WikiRik (approved):

PR #1923, opened 2026-03-06
fix(statutory): restrict post-deadline application body reassignment

WikiRik:
## Summary
- Require `manage_applications` when changing an application's `body_id` after `application_period_ends`.
- Keep the membership guard for body changes so reassignment is only allowed to bodies the applicant belongs to.
- Add API tests covering allowed/forbidden post-deadline body changes for apply vs manage permissions.

## Testing
- `NODE_ENV=test npm run db:setup && npx jest test/api/applications-editing.test.js --runInBand --forceExit`
- `npx eslint lib/applications.js test/api/applications-editing.test.js`

### DIFF
--- a/statutory/lib/applications.js
+++ b/statutory/lib/applications.js
@@ -305,7 +305,12 @@ exports.updateApplication = async (req, res) => {
     }
 
     const user = await core.getMember(req, req.application.user_id);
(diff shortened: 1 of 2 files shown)


WikiRik:
No need to review this. I'm not sure if this is the way I actually want to go. The application deadline is for sure not important, but I think we need a broader approach to ensure that Chair and admins are able to change all applications

PR #1942, opened 2026-03-15
feat(infra): switch to pg_dump custom format and add volume backups

WikiRik:
## Summary

- **dump.sh**: Switches from `pg_dumpall` (text format) to `pg_dump -Fc` (custom format) with per-database dumps. Custom format is compressed by default and supports parallel restore via `pg_restore -j`.
- **dump.sh**: Adds docker volume backup support — pass volumes after `--` separator (e.g., `dump.sh postgres-core -- core-media events-media`).
- **restore.sh**: Uses `pg_restore -j 4` for parallel restore from custom format dumps (mounted as files, not piped via stdin which corrupts binary format).
- **restore.sh**: Auto-detects backup format — falls back to legacy `psql` restore for old text-format `pg_dumpall` backups.
- **restore.sh**: Auto-detects and restores volume backups (`.tar` files) found in the archive, creating missing volumes if needed.

## Test plan

- [x] Verified `pg_dump -Fc` and `pg_restore -j 4` flags are available in PostgreSQL 10.21
- [x] Tested dump: created test postgres container, seeded data, dumped with `-Fc` (without `-t` flag to avoid tty binary corruption)
- [x] Tested restore: dropped database, restored from custom format dump with `pg_restore -j 4` via file mount — all data recovered
- [x] Tested volume backup: created test volume, backed up to tar, verified contents
- [x] Tested volume restore: cleared volume, restored from tar, verified all files recovered

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/scripts-server/dump.sh
+++ b/scripts-server/dump.sh
@@ -1,7 +1,12 @@
 #!/bin/bash
 
-# Usage: dump.sh <name of the postgres hosts to backup>
+# Usage: dump.sh <name of the postgres hosts to backup> [-- <docker volumes to backup>]
 # The backup location is /opt/backups
+#
+# Example: dump.sh postgres-core postgres-events -- core-media events-media
+#
+# Uses pg_dump in custom format (-Fc) for each database, which is compressed
+# and supports parallel restore via pg_restore -j.
 
 # Idea of this script: Spawn a container which links up to any db, does the
 #   dump through the network and then save the dump outside of the container.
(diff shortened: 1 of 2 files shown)


WikiRik:
@linuxbandit could you take a look at this soon?

scripts-server/dump.sh
```diff
@@ -1,7 +1,12 @@
 #!/bin/bash
 
-# Usage: dump.sh <name of the postgres hosts to backup>
+# Usage: dump.sh <name of the postgres hosts to backup> [-- <docker volumes to backup>]
```
linuxbandit, line 3: I'd not mix postgres and volumes backup, i would have dump.sh being a 'parent' script that runs e.g. dump-pg.sh and dump-volume.sh

scripts-server/restore.sh
```diff
@@ -41,79 +40,199 @@ tar --force-local --one-top-level="${tmp_folder}" -xvf "${input_file}"
...
+#shellcheck disable=SC2207
+dump_files=( $(find "${tmp_folder}" -name "postgres-*-*.dump" -type f 2>/dev/null) )
 
-  if [[ $(wc -c <"${file}") -le 100 ]]
-  then
-    echo "Something went wrong replaying ${name}, file ${file} is empty"
-    error=1
-  fi
+if [[ ${#dump_files[@]} -gt 0 ]]; then
+  echo "Detected custom format backup (.dump files) - using pg_restore with parallel restore"
```
linuxbandit, line 50: not sure i am sold about a custom format, but i admittedly did not understand whether this is just a custom extension

linuxbandit (changes requested):
I think we better work on this together next sunday, so that we have at least one solution which we will use to backup ("in the new way") the DBs. I will also share with you the script I use in my homelab (perhaps I will port it to the ansible), but it basically stops all the services at 3am to back them up and then restarts them. This to avoid write issues during the backup phase, and therefore corruptions

linuxbandit:
love this btw and sorry, didn't see the notification (we have too many dependabot PRs...!)

btw about dependabot, a 'radical' take: https://words.filippo.io/dependabot/

PR #1978, opened 2026-04-15
fix(statutory): expose all applicable event bodies

WikiRik:
## Summary
- compute statutory `apply_from_body` from all bodies returned by Core instead of only the logged-in user's memberships
- fix the manager application edit flow so body options are not reduced to the overlap with the editor's own bodies
- add a regression test covering an event body that is outside the current user's memberships

## Verification
- `NODE_ENV=test npm run db:setup`
- `NODE_ENV=test jest test/api/events-display.test.js test/api/applications-editing.test.js --runInBand --forceExit`


WikiRik:
Superseded by #1979

PR #1985, opened 2026-04-19
chore(monorepo): remove gsuite-wrapper and Redis tooling

WikiRik:
## Summary
- remove the `gsuite-wrapper` workspace and delete its source, tests, CI jobs, Dependabot entries, and monorepo metadata
- remove frontend, docs, and helper-script references to the retired service so the project no longer advertises or targets it
- remove the remaining Redis admin/config wiring because Redis was only being kept for `gsuite-wrapper`

## Testing
- not run


core/scripts/sync-gsuite-users.js
CASician, line 1: Do we want to replicate this in the future? It doesn't look completely obsolete. 
But, on the other hand, I will most probably create it from scratch anyway. I might keep this file in mind for reference.

frontend/src/views/static/About.vue
```diff
@@ -168,11 +168,6 @@
                 <td><a href="https://github.com/AEGEE/network">network</a></td>
                 <td>Network history and status.</td>
               </tr>
-
-              <tr>
-                <td><a href="https://github.com/AEGEE/gsuite-wrapper">gsuite-wrapper</a></td>
```
CASician, line 173: We will still have the module. Just change the url? Actually, the urls refer to the old repos (before monorepo), so general cleaning of the file would be nice.

Makefile
CASician, line 1: Was this backup working? Does it have any relevant data?

CASician (approved):

WikiRik:
@CASician thanks for the feedback! I'll go through them and the changes in general before I merge

Makefile
WikiRik: On line 156 it says; `echo "TODO: redis"` so no

core/scripts/sync-gsuite-users.js
WikiRik: We still have it in the earlier commits, so it's not lost. Creating from scratch to have a regular sync as backup is nice

frontend/src/views/static/About.vue
WikiRik: Good point, we'll fix that later. We also haven't archived the old repos since some PRs are still open on there

PR #1992, opened 2026-04-26
fix(frontend): add instructions for contacts on UpdateProfile

WikiRik:


frontend/src/views/core/members/UpdateProfile.vue
```diff
@@ -98,11 +98,11 @@
...
-      <div class="field" v-if="validationErrors.bodies">
-        <label class="label">
-          Body <span class="has-text-danger">*</span>
-          <tooltip text="You can change your notification email in your profile" />
-        </label>
+        <div class="field" v-if="validationErrors.bodies">
+          <label class="label">
+            Body <span class="has-text-danger">*</span>
+            <tooltip text="You can change your notification email in your profile" />
+          </label>
```
WikiRik:
```suggestion
      <div class="field" v-if="validationErrors.bodies">
        <label class="label">
          Body <span class="has-text-danger">*</span>
          <tooltip text="You can change your notification email in your profile" />
        </label>
```

PR #2158, opened 2026-09-07
fix(statutory): base antenna quorum on publication and acceptance

WikiRik:
Statutory quorum statistics now count antennae only and display the 50% requirement, rounded up. Previously the chart also included contact antennae and partners.

Before the participant-list publish time, count antennae with any non-cancelled application. At and after that time, count antennae with an accepted non-cancelled application. The server uses participants_list_publish_deadline, regardless of staff permissions or early status reveal, and returns the filtered body IDs plus the basis used. Other application statistics retain their existing meaning.

The UI labels the current basis and does not state whether quorum is reached, since other determining factors are not represented in the system. Empty antenna lists produce valid percentages.

CI now discovers test/**/*.js recursively across all six Jest service jobs, including unit tests previously excluded by API-only globs. The glob is quoted so CircleCI performs recursive expansion.

Validation: 8 Jest unit tests passed, covering the publication boundary, time-zone offsets, duplicate bodies, cancelled applications, early status reveal, default server time, and empty results. The quorum helper has 100% statement, function, and branch coverage. Recursive discovery matches all 240 suites found by unrestricted Jest discovery across the six services. YAML and shell syntax checks passed. Four additional server-to-chart calculation scenarios passed. ESLint passed for all changed JavaScript/Vue files, and git diff --check passed. Full database/API and browser tests were not run.

Deployment: deploy the statutory service before the frontend, which consumes the new quorum response field.

### DIFF
--- a/statutory/README.md
+++ b/statutory/README.md
@@ -24,6 +24,20 @@ The main developer of this module is Sergey Peshkov (AEGEE-Voronezh, github.com/
 
 You can specify the microservice configuration by editing the `lib/index.js` file. Check out the example at `lib/index.js.example` and the comments in `lib/index.js` for more information.
 
+## Quorum statistics
+
+`GET /events/:event_id/applications/stats` includes `data.quorum` with
+`accepted_only` (boolean) and `body_ids` (distinct represented body IDs).
+Before `participants_list_publish_deadline`, representation includes all
+non-cancelled applications. At and after that timestamp, it includes only
+accepted non-cancelled applications. The cutoff uses server time, regardless
+of staff permissions or early application-status publication. Other statistics,
+including `by_body`, retain their existing meaning.
+
+The frontend restricts these IDs to antennae and displays a 50% requirement,
+rounded up. These statistics do not determine whether statutory quorum is met.
+Deploy the updated statutory service before the updated frontend.
+
 ## LICENSE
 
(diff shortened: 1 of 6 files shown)


statutory/README.md
```diff
@@ -24,6 +24,20 @@ The main developer of this module is Sergey Peshkov (AEGEE-Voronezh, github.com/
...
+Before `participants_list_publish_deadline`, representation includes all
+non-cancelled applications. At and after that timestamp, it includes only
+accepted non-cancelled applications. The cutoff uses server time, regardless
+of staff permissions or early application-status publication. Other statistics,
+including `by_body`, retain their existing meaning.
+
+The frontend restricts these IDs to antennae and displays a 50% requirement,
+rounded up. These statistics do not determine whether statutory quorum is met.
+Deploy the updated statutory service before the updated frontend.
+
```
WikiRik:
```suggestion
```

statutory/lib/quorum.js
```diff
@@ -0,0 +1,12 @@
+// Quorum representation follows the scheduled publication time, regardless of
+// staff permissions or an early application-status reveal.
```
WikiRik:
```suggestion
// permissions or an early application-status reveal.
```

